PR #2, opened 2017-09-08
Pass query strings through proxy

rschroll:
In my testing, query string were being dropped, instead of being sent through to the proxy.  This change fixes things for me.

It's probably more complicated than strictly necessary (we could append '?' when the query string is empty), but it should keep the URLs as close as possible to their original form.


ryanlovett:
Merged, thanks!

PR #7, opened 2017-10-31
Support a lot more HTTP methods

### LINKED ISSUE #5: Support all HTTP methods
I think it's ok for us to support all HTTP methods, rather than just POST and GET

yuvipanda:
I don't think there's any explicit reason for us to whitelist
methods.

Fixes #5


yuvipanda:
@ryanlovett can you make a release or add me as a maintainer on pypi after this gets merged? Thanks!

ryanlovett:
I updated setup.py after your previous patch, but you sent another PR before I could build. :)

I'll just add you on pypi.

PR #8, opened 2017-11-03
Fix typo causing query params to fail

yuvipanda:


yuvipanda:
@ryanlovett can you also make a release after this?

ryanlovett:
@yuvipanda Done and done. Thanks!

PR #11, opened 2018-01-05
Add support for proxying websockets too

### LINKED ISSUE #6: Support websockets
Right now we do not support websockets.

We should!

yuvipanda:
Unfortunately this also bumps up minimum required Python version to 3.5
in the same commit. I'm ok with that, but should've ideally been a separate PR.  
  
Fixes #6

### DIFF
--- a/nbserverproxy/handlers.py
+++ b/nbserverproxy/handlers.py
@@ -5,18 +5,67 @@ Some original inspiration from https://github.com/senko/tornado-proxy
 """
 import socket
 import os
-from tornado import gen, web, httpclient, httputil, process
+from tornado import gen, web, httpclient, httputil, process, websocket
 
 from notebook.utils import url_path_join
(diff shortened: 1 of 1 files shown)


nbserverproxy/handlers.py
```diff
@@ -5,18 +5,88 @@
...
+        Called when the client closes our websocket connection.
+
+        We close our connection to the backend too.
+        """
+        self.ws.close()
 
     @web.authenticated
-    @gen.coroutine
-    def proxy(self, port, proxied_path):
+    async def proxy(self, port, proxied_path):
```
betatim: what does this change to do the minimum python version required?

minrk: makes it Python 3.5

yuvipanda: I put that info in the commit message, forgot to put it in PR. Have done so now!

PR #14, opened 2018-01-17
Do not use None as body for POST requests

betatim:
POST requests can not have None as body, so instead use an empty byte sequence.

I am actually not sure if this is the "right way" (tm) to fix this. Or what else this potentially breaks in terms of HTTP compliance.

Needs this to make https://github.com/betatim/openrefineder work. I think the real problem is that OpenRefine sends empty bodys for POST requests???


ryanlovett:
There's some discussion of this at https://github.com/tornadoweb/tornado/issues/1213. Seems like a valid thing to do.

PR #15, opened 2018-01-20
[MRG] Use MRO method to get base classes

betatim:
Using the `__bases__` attribute stops working if you have a class that inherits from a class that uses the `WebSocketMixin`. Instead use the `inspect` module to get the classes in the ["method resolution order"](https://docs.python.org/3/library/inspect.html#inspect.getmro).

I think this should be more robust. Though I am not sure why we do this trickery, seems a bit weird/smells wrong.

* [x] Before merging this we should move the import to the top.


betatim:
Ready to merge

PR #17, opened 2018-01-25
Fix asynchronity issues with websockets

yuvipanda:
It looks like Tornado's websocket support is a little weird,
in that the client is asynchronous but the server seems synchronous?

That's a little strange, so we had to do a bunch of ioloop stuff
around this.

Also adds support to downstream SuperviceAndProxy handler for
using websockets properly


yuvipanda:
This thing needs tests lol.

yuvipanda:
(but we shouldn't have tests block this PR, since master is broken without it)

PR #18, opened 2018-02-09
Update README.md

lsetiawan:
## Overview

Add instruction to install from conda and tweaked instruction a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,15 +12,7 @@ The service is proxied off of `/proxy/` at https://address.of.notebook.server/us
 ## Installation
 To install:
 ```
-pip install git+https://github.com/jupyterhub/nbserverproxy
-```
-or 
-```
 conda install -c conda-forge nbserverproxy
-```
-Enable extension by running:
-```
-jupyter serverextension enable --py nbserverproxy
 ```
 
 Based on https://github.com/senko/tornado-proxy.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,9 +10,16 @@ python -m http.server 8001
...
-Run
+To install:
 ```
 pip install git+https://github.com/jupyterhub/nbserverproxy
+```
+or 
+```
+conda install -c conda-forge nbserverproxy
+```
+Enable extension by running:
```
betatim: I've seen other extensions that did not need enabling anymore when they were installed via conda. Do you know what needs to happen on the conda package side for that to be the case here as well?

betatim:
Thanks for contributing and looks good to me. Let's figure out the enabling of extensions bit and then merge.

README.md
lsetiawan: I'm working on enabling this on the conda side. Stay tuned. Thanks.

lsetiawan:
Sorry all for the late update: But I've tested the extensions via conda install, and all seems to work. Thanks.

https://anaconda.org/conda-forge/nbserverproxy

betatim:
What was your thinking behind removing the `pip` based install instructions? I think keeping them would be useful for those who don't have conda to remind them that that they have to do more than just `pip install`.

lsetiawan:
@betatim Apologies. 

> I think keeping them would be useful for those who don't have conda to remind them that that they have to do more than just `pip install`.

Yea that makes sense. I've added it back. Thanks. Hope that works.

betatim:
No worries :) This is why we talk about things.

Merging now and congratulations on your first contribution to this project!

PR #23, opened 2018-03-06
record proxied activity

minrk:
store it in api_last_activity so that the notebook server’s activity monitor will recognize proxied activity


ryanlovett:
Thanks!

PR #26, opened 2018-03-27
Add support for Sec-WebSocket-Protocol.

ryanlovett:


ryanlovett:
@yuvipanda or @betatim: if you can review this that'd be great. Otherwise I'll merge in a week or so.

betatim:
Do you have an example I can try it out with? My websocket-ish is a bit rusty but I could review it by using it :)

ryanlovett:
The only one I have is the one I developed this feature for. :)

https://github.com/ryanlovett/binder-novnc

You can view it on binder at https://mybinder.org/v2/gh/ryanlovett/binder-novnc/master?filepath=start_vnc.ipynb. You execute the cells then click on the link it generates. In this case, novnc looks for either the "binary" or "base64" subprotocols.

Thanks for taking a look!

ryanlovett:
I think I found a more general way of doing this that's probably better. Basically just pass along all of the headers instead of just the ones we look for. Will test.

Issue #1, opened 2017-03-31
Make it a standalone service


minrk:
It would be useful to have this as a standalone service, to enable deploying any web application behind 
JupyterHub auth, not just as a notebook extension.

The package could still provide a serverextension for the existing case.

BerserkerTroll:
@minrk this seem to work for me
```python
#!/usr/bin/python3


import getopt
import json
import pprint
import os
import sys
from urllib.parse import urlparse

import requests
from tornado import web, httpserver, ioloop

from jupyterhub.services.auth import HubOAuthenticated, HubOAuthCallbackHandler

import nbserverproxy


class APIHandler(web.RequestHandler):
    """Relay API requests to the Hub's API using the service's API token."""
    def get(self, path):
        api_token = os.environ['JUPYTERHUB_API_TOKEN']
        api_url = os.environ['JUPYTERHUB_API_URL']
        r = requests.get(api_url + '/' + path,
            headers={'Authorization': 'token %s' % api_token},
        )
        r.raise_for_status()
        self.set_header('Content-Type', 'application/json')
        self.write(r.text)


class EchoHandler(HubOAuthenticated, web.RequestHandler):
    """Reply to an HTTP request with the path of the request."""
    @web.authenticated
    def get(self):
        self.write(self.request.path)


def main():
    pprint.pprint(dict(os.environ), stream=sys.stderr)

    StandaloneLocalProxyHandler = type(
        'StandaloneLocalProxyHandler',
        (HubOAuthenticated, nbserverproxy.handlers.WebSocketHandlerMixin, web.RequestHandler),
        dict(nbserverproxy.handlers.LocalProxyHandler.__dict__)
    )

    app = web.Application([
        (r'.*/api/(.*)', APIHandler),
        (os.getenv('JUPYTERHUB_OAUTH_CALLBACK_URL'), HubOAuthCallbackHandler),
        (os.getenv('JUPYTERHUB_SERVICE_PREFIX') + r'proxy/(\d+)(.*)', StandaloneLocalProxyHandler),
        (r'.*', EchoHandler),
    ], cookie_secret=os.urandom(32))

    optlist, args = getopt.getopt(sys.argv[1:], '', ['port='])
    port = dict(optlist)['--port']

    server = httpserver.HTTPServer(app)
    server.listen(port)
    try:
        ioloop.IOLoop.instance().start()
    except KeyboardInterrupt:
        print('\nInterrupted')


if __name__ == '__main__':
[...]

PR #38, opened 2018-06-05
Correct WebSocket proxying

### LINKED ISSUE #35: Proxying Xpra is not working
I've tried to proxy [Xpra](https://www.xpra.org/) [html5 client](https://xpra.org/trac/wiki/Clients/HTML5) with `nbserverproxy` (version 0.8.3), but it doesn't seem to work.
When I start Xpra on a local machine with the
```
xpra start --bind-tcp=0.0.0.0:14500 --html=on --start=xterm
```
command and open `localhost:14500`, I see `xterm` in my browser.

I've tried to run this command from the `jupyter` terminal and then access `.../proxy/14500`.
First of all, the HTML page tries to load `js` (or `css`) from `.../proxy/js/` instead of `.../proxy/14500/js`. Ok, I've added a slash to the url and then the page was able to load.

But then, probably because it can't connect to websockets, the [`index.html`](https://github.com/totaam/xpra-html5/blob/master/index.html) page redirects me to [`connect.html`](https://github.com/totaam/xpra-html5/blob/master/connect.html) page, which locally happens only when I stop X server. As I said, this likely means that it has problems with websockets so it thinks that the server is down. **UPD**: In Firefox Network console, the last request from `index.html` page I see is websockets request to the `/`, which is expectedly redirected to `/hub`.

There is Wiki page about proxying Xpra with Nginx: https://www.xpra.org/trac/wiki/Nginx
AFAIU it makes Nginx to do something to HTTP headers so that websockets work behind a proxy.

I know about https://github.com/ryanlovett/nbnovnc/, but I would like to see Xpra working.

BerserkerTroll:
Subject to careful review. Not sure where to put `self._record_activity()` and whether `self.set_status(400)` is enough to send `400 Bad Request` back to the client in case the backend does not support the WebSocket (or any other error happens).

Fixes #35

### DIFF
--- a/nbserverproxy/handlers.py
+++ b/nbserverproxy/handlers.py
@@ -93,12 +93,11 @@ class WebSocketHandlerMixin(websocket.WebSocketHandler):
     async def get(self, *args, **kwargs):
         if self.request.headers.get("Upgrade", "").lower() != 'websocket':
             return await self.http_get(*args, **kwargs)
-        # super get is not async
-        super().get(*args, **kwargs)
+        return await self.ws_get(*args, **kwargs)
 
 
 class LocalProxyHandler(WebSocketHandlerMixin, IPythonHandler):
-    async def open(self, port, proxied_path=''):
+    async def ws_get(self, port, proxied_path=''):
         """
         Called when a client opens a websocket connection.
 
@@ -144,12 +143,18 @@ class LocalProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             self.log.info('Trying to establish websocket connection to {}'.format(client_uri))
             self._record_activity()
             request = httpclient.HTTPRequest(url=client_uri, headers=headers)
-            self.ws = await pingable_ws_connect(request=request,
-                on_message_callback=message_cb, on_ping_callback=ping_cb)
-            self._record_activity()
-            self.log.info('Websocket connection established to {}'.format(client_uri))
+            try:
+                self.ws = await pingable_ws_connect(request=request,
+                    on_message_callback=message_cb, on_ping_callback=ping_cb)
+            except Exception:
+                self.set_status(400)
+            else:
+                self._record_activity()
+                self.log.info('Websocket connection established to {}'.format(client_uri))
+                # super get is not async
+                super(WebSocketHandlerMixin, self).get(port, proxied_path)
 
-        ioloop.IOLoop.current().add_callback(start_websocket_connection)
+        return await start_websocket_connection()
 
     def on_message(self, message):
         """


ryanlovett:
I'm leary about abandoning the open() call. According [to spec](https://tools.ietf.org/html/rfc6455#section-4.1) clients must open the connection as part of the handshake so if we don't open() the proxied service when the client calls open() on us, some clients might see problems. That being said, this patch worked for you so I need to do more research.

BerserkerTroll:
@ryanlovett I think you misinterpret what `WebSocketHandler.open` is and when it is called. It is not called when a client opens a connection to a server (before GET Upgrade request is sent over this connection), it is called only after `WebSocketHandler` responds to Upgrade request!

This is done in `WebSocketProtocol13` class, which is instantiated by `WebSocketHandler`.
From `WebSocketHandler.get`:
```python
        self.ws_connection = self.get_websocket_protocol()
        if self.ws_connection:
            self.ws_connection.accept_connection()
```
and `get_websocket_protocol` is defined as:
```python
    def get_websocket_protocol(self):
        websocket_version = self.request.headers.get("Sec-WebSocket-Version")
        if websocket_version in ("7", "8", "13"):
            return WebSocketProtocol13(
                self, compression_options=self.get_compression_options())
#               ^^^^ passes self to WebSocketProtocol13, which is accepted as a `handler` in that class
```
`WebSocketProtocol13.accept_connection` invokes `WebSocketProtocol13._accept_connection`. Which answers to Upgrade request and calls the `open` callback:
```python
        self.handler.set_status(101)
        self.handler.set_header("Upgrade", "websocket")
        self.handler.set_header("Connection", "Upgrade")
        self.handler.set_header("Sec-WebSocket-Accept", self._challenge_response())
        self.handler.finish()

        self.handler._attach_stream()
        self.stream = self.handler.stream

        self.start_pinging()
        self._run_callback(self.handler.open, *self.handler.open_args,
                           **self.handler.open_kwargs)
```

ryanlovett:
Yes, I had been thinking that open() was called when a client connects to the server. Thanks for the additional insight! Let me go over this a bit more.

If this is time-critical for you I can try to escalate -- just let me know.

BerserkerTroll:
>If this is time-critical for you I can try to escalate -- just let me know.

No problem, if I needed it right now I'd use my fork. Anyway, for what I want to do with this proxy I need more changes than this bugfix.

BerserkerTroll:
>  this pretty much broke nbnovnc

The implementation of ws proxying is broken. If something like nbnovnc works it is either a pure luck or it was written to work with broken ws proxying and thus does not work with correct ws proxying.

>  and relies on its get to mix ws and http connections in one handler

`get` still mixes ws and http connections in one handler. It even does it more "directly".

> What bug is this PR addressing btw?

The OP message explicitly says what this PR fixes.

adsche:
> > this pretty much broke nbnovnc
> 
> The implementation of ws proxying is broken. If something like nbnovnc works it is either a pure luck or it was written to work with broken ws proxying 

I don't believe this is true. nbnovn is really simple (1 file, <100 LOC), the only relevant overloaded function would be https://github.com/ryanlovett/nbnovnc/blob/master/nbnovnc/handlers.py#L88 and that is not working around anything.

I tried two echo servers, both work fine (after #51 also binary). I can also point nbserverproxy directly to a running websockify instance and it works just fine (noVNC and any other TCP socket, can easily be tested with nc). 

> and thus does not work with correct ws proxying.

In the same way, I can point my reverse proxy to those endpoints and it works no different than with nbserverproxy.

> The OP message explicitly says what this PR fixes.

Yup, sorry about that. Would love to test xpra myself, but a dependency is not packaged for my platform.

Anyway, I'm out of here now. Good luck!

BerserkerTroll:
@adsche  `SuperviseAndProxyHandler` overrides the `open` function, which was used incorrectly, see https://github.com/jupyterhub/nbserverproxy/pull/38#issuecomment-398321288
This is an example of how something was written to work with incorrect ws proxying implementation.

I'll add the fix for `SuperviseAndProxyHandler`. Or just make another PR fixing #1 and close this one.

manics:
I'll close this since the author of this PR has deleted their account.

PR #46, opened 2018-08-15
Allow users to connect to services on other hosts

jacobtomlinson:
I've come across situations where it would be useful to proxy to a port on a host other than `localhost`. 

This PR adds an optional hostname option to the proxy url so users can specify things like:

```
/user/jacobtomlinson/lab/proxy/example.com/8080/somepath.html
```

The original format is still valid.

```
/user/jacobtomlinson/lab/proxy/8080/somepath.html
```

### DIFF
--- a/nbserverproxy/handlers.py
+++ b/nbserverproxy/handlers.py
@@ -99,7 +99,7 @@ class WebSocketHandlerMixin(websocket.WebSocketHandler):
 
 
 class LocalProxyHandler(WebSocketHandlerMixin, IPythonHandler):
-    async def open(self, host='localhost', port, proxied_path=''):
+    async def open(self, host, port, proxied_path=''):
         """
         Called when a client opens a websocket connection.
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
@yuvipanda that sounds sensible! How do you recommend implementing this?

Two options I can think of are:
 - Additional Jupyter config options specific in extra config
 - An environment variable

yuvipanda:
I'd suggest using traitlets backed by environment variables. github.com/jupyterhub/nbgitpuller is an example of how to do this.

rcthomas:
I confirm the awesome.  I just tested this out with Dask and it was success, with a tiny tweak.  In our case we want to get to an IP, not a domain name.  I swapped out the domain name for a really bad IP regex (I can make it better) and it worked.  Could we make it handle IPs instead of just domain names too?

manics:
Closed by https://github.com/jupyterhub/jupyter-server-proxy/pull/154

PR #51, opened 2018-10-18
Make sure binary messages are proxied correctly (fixes #50)

### LINKED ISSUE #50: Binary websockets currently not working (easy fix)
Hi,

first of all, thanks for the great project, I was halfway through implementing something similar myself when I found it.

I'm using it to proxy a VNC connection (like nbnovnc) which uses a binary websocket. I found that connection could be established but traffic would stop after the RFB handshake (which just states the protocol version in all ASCII characters).

Turns out, binary messages from the backend are not relayed properly to the frontend. The reason is https://github.com/jupyterhub/nbserverproxy/blob/master/nbserverproxy/handlers.py#L163

Similar to when proxying from frontend to backend, it should instead be:

    self.ws.write_message(message, binary=type(message) is bytes)

With that, e.g., nbnovnc runs perfectly again.

Cheers!

adsche:
- in on_message(), check if the received message is of type bytes and if so, send it on with binary flag
- change 'message is bytes' to isinstance(message, bytes)

### DIFF
--- a/nbserverproxy/handlers.py
+++ b/nbserverproxy/handlers.py
@@ -130,7 +130,7 @@ class LocalProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             if message is None:
                 self.close()
             else:
-                self.write_message(message, binary=type(message) is bytes)
+                self.write_message(message, binary=isinstance(message, bytes))
 
         def ping_cb(data):
             """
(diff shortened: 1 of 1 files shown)


betatim:
Thanks for this! CLoses #50

PR #52, opened 2018-10-21
record api_last_activity as timezone aware timestamp

gweis:
fixes problems with notebook status api when producing last_activity timestamp.

i.e. in this line https://github.com/jupyter/notebook/blob/4f9a6cbdd81ed26a7979d993766d7223e1c72e5c/notebook/notebookapp.py#L367
when comparing (max(...)) all timestamps it produces a
TypeError: can't compare offset-naive and offset-aware datetimes


betatim:
Thanks! I'll merge this now and if there is a problem we can follow up in a new PR.

PR #56, opened 2018-11-19
Allow for additional, or alternative, options for proxy request

nokome:
This change allows handler classes that extend `LocalProxyHandler` to specify additional, or alternative, arguments to `tornado.httpclient.HTTPRequest` when proxying. 

For example, this was needed so that we could increase the `request_timeout` in https://github.com/minrk/nbstencilaproxy to handle the execution of long running cells.


ryanlovett:
Thanks!

PR #59, opened 2018-12-10
Convert port number to string before building URL

betatim:
I think this will fix the problem we are seeing on mybinder.org with newly built repositories that use RStudio (xref https://github.com/jupyterhub/mybinder.org-deploy/issues/865).


betatim:
To fix the problem on mybinder.org we also need to cut a new release. I think this would be 0.8.7 (from PyPI).

PR #64, opened 2018-12-28
Use simpervisor for process supervision

yuvipanda:
The [simpervisor library](http://github.com/yuvipanda/simpervisor) 
handles process supervision properly, including:
    
- Restarting processes when they die
- Readyness checks
- Passing signals to child processes appropriately

We also use a lock to prevent concurrent process starts,
rather than boolean variables.


yuvipanda:
That's done now. This is ready for review / merge.

PR #65, opened 2018-12-30
Allow configuring processes to start be much easier

yuvipanda:
- Traitlet based configuration for users to directly
  specify what processes they want to start & proxy
- entrypoint based configuration for users to be able
  to install a package that can specify how processes
  should start


ryanlovett:
Woot!

PR #79, opened 2019-01-04
base_url and timeout server config

manics:
Adds a timeout parameter for the readiness check, this is needed for applications that are slow to start. The default is `5s` which matches the current timeout from https://github.com/yuvipanda/simpervisor/blob/b736f117f30ab59db7fbd2ae7637b9fdf6d0295d/simpervisor/process.py#L19

Adds `{base_url}` as a template parameter for configuring servers. This is useful for applications such as many Django apps which need to know their visible path.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -316,7 +316,9 @@ class SuperviseAndProxyHandler(LocalProxyHandler):
                 # Set up extra environment variables for process
                 server_env.update(self.get_env())
 
-                proc = SupervisedProcess(self.name, *cmd, env=server_env, ready_func=self._http_ready_func, log=self.log)
+                timeout = self.get_timeout()
+
+                proc = SupervisedProcess(self.name, *cmd, env=server_env, ready_func=self._http_ready_func, ready_timeout=timeout, log=self.log)
                 self.state['proc'] = proc
 
                 try:
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -287,6 +287,12 @@ class SuperviseAndProxyHandler(LocalProxyHandler):
            overridden in subclasses.'''
         return {}
 
+    def get_timeout(self):
+        """
+        Return timeout (in s) to wait before giving up on process readiness
+        """
+        return 5
+
     async def _http_ready_func(self, p):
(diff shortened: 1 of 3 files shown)


yuvipanda:
<3 this is great! The one issue here is that since we are passing an extra parameter (port) to all functions, any already existing functions used will now fail.

The way around it is to *only* pass parameters that the functions declare. We can do something like

```python
asked_args = callback.code.co_varnames
passed_args = {k: v for k, v in all_args if k in asked_args}
callback(**passed_args)
```

This makes it so that timeout is passed to the callback function *only* if the function asks for it. Extremely simple 'dependency injection' of sorts.

How do you feel about adding that in this PR? Would be lovely & highly appreciated.

Thanks!

yuvipanda:
I've a PR coming up that should implement the 'dependency injection'

betatim:
> The way around it is to _only_ pass parameters that the functions declare. We can do something like
> 
> ```python
> asked_args = callback.code.co_varnames
> passed_args = {k: v for k, v in all_args if k in asked_args}
> callback(**passed_args)
> ```
> 

https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object could be interesting for off-loading the "digging around the internals of python internals" bit.

yuvipanda:
Apologies for the delay, @manics. I've pushed up my work as https://github.com/jupyterhub/jupyter-server-proxy/pull/82 - pretty similar to yours, except I also added a test and a teensy bit more docs. I'm happy to rebase my work on yours too if you'd prefer that instead.

jupyter_server_proxy/handlers.py
```diff
@@ -316,7 +316,9 @@ def get_env(self):
                 # Set up extra environment variables for process
                 server_env.update(self.get_env())
 
-                proc = SupervisedProcess(self.name, *cmd, env=server_env, ready_func=self._http_ready_func, log=self.log)
+                timeout = self.get_timeout()
```
yuvipanda: Can you provide a default implementation of this too, since people might wanna use this handler by itself?

yuvipanda:
@manics one minor comment, LGTM otherwise.

yuvipanda:
@manics I've just made that change, will merge shortly

manics:
@yuvipanda Sorry for being slow, I'm at a workshop.

PR #82, opened 2019-01-08
Call command / environment with only args they ask for

yuvipanda:
This lets us add more parameters in the future without
breaking backwards compatibility

### DIFF
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -91,7 +91,7 @@ with:
        'command': _cmd_callback
    }
 
-The ``port`` argument will be pased to the callable. This is a simple form of dependency
+The ``port`` argument will be passed to the callable. This is a simple form of dependency
 injection that helps us add more parameters in the future without breaking backwards
 compatibility.
 
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -26,8 +26,8 @@ pairs.
      process. If the string ``{port}`` is present anywhere, it'll
      be replaced with the port the process should listen on.
     
-   * A callable that takes one parameter - ``port``, and returns a
-     list of strings that are used & treated same as above.
+   * A callable that takes any :ref:`callable arguments <server-process/callable-argument>`,
+     and returns a list of strings that are used & treated same as above.
   
    This key is required.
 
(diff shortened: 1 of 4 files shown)


docs/server-process.rst
```diff
@@ -60,6 +60,60 @@ pairs.
...
+.. code:: python
+
+   def _cmd_callback(port):
+       return ['some-command', '--port=' + str(port)]
+
+   server_config = {
+       'command': _cmd_callback
+   }
+
+The ``port`` argument will be pased to the callable. This is a simple form of dependency
```
ryanlovett: s/pased/passed/

docs/server-process.rst
```diff
@@ -60,6 +60,60 @@ pairs.
...
+#. **port**
+   The port the command should listen on
+
+If any of the returned strings, lists or dictionaries contain strings
+of form ``{<argument-name>}``, they will be replaced with the value
+of the argument. For example, if your function is:
+
+.. code:: python
+
+   def _openrefine_cmd():
```
ryanlovett: Does this need to be
```suggestion
   def _openrefine_cmd(port):
```

jupyter_server_proxy/utils.py
```diff
@@ -0,0 +1,32 @@
+def call_with_asked_args(callback, args):
+    """
+    Call callback only the args it wants from args
```
ryanlovett:
```suggestion
    Call callback with only the args it wants from args
```

docs/server-process.rst
yuvipanda: It doesn't! We don't need to use port, so we don't have to ask for it.

yuvipanda:
@ryanlovett thanks for the review! I've fixed them now

jupyter_server_proxy/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    # We only pick the local variables
+    asked_arg_names = callback.__code__.co_varnames[:callback.__code__.co_argcount]
+    asked_arg_values = []
+    missing_args = []
+    for asked_arg_name in asked_arg_names:
+        if asked_arg_name in args:
+            asked_arg_values.append(args[asked_arg_name])
+        else:
+            missing_args.append(asked_arg_name)
+    print(asked_arg_values)
```
manics: Is this print statement intentional?

yuvipanda:
@manics I've also invited you to this repo as a collaborator, since I think you are also very interested in seeing this move forward.

Issue #43, opened 2018-07-23
Doesn't work with applications which use absolute URLs


jacobtomlinson:
I've noticed that a few applications that I've tried to proxy recently use absolute urls.

For example when running on a regular server if you go to `http://localhost:1234/` it will look for a stylesheet at`/style/index.css`. This works fine when running locally but when using nbserverproxy I visit `http://localhost:8888/lab/proxy/1234` which still looks for styles at `/style/index.css` rather than `/lab/proxy/1234/style/index.css`.

These applications often take a `baseurl` config option somewhere which lets you prepend all urls with a base path, such as `/lab/proxy/1234`. This would allow me to fix this functionality. However it appears nbserverproxy rewrites its urls so when I visit `http://localhost:8888/lab/proxy/1234` the path gets rewritten to `/`. This then causes problems because the application is now expecting a base url of `/lab/proxy/1234/`. This results in the HTML page not displaying because flask or whatever doesn't have a route for `/`, but the CSS would display ok if it were ever requested.

betatim:
No real answers, just a short clarification: if you visit `http://localhost:8888/lab/proxy/1234/` (with trailing slash) does that change things at all? It is starting to feel like 'have you tried with trailing slash' should become a FAQ.

manics:
I ran into this last week with a django app (with baseurl=`/proxy/80`, trailing slash in config or accessed URL didn't help). In this particular case it was fronted by Nginx so I could work around by adding a rewrite config: `location / { rewrite ^/(.*)$ /proxy/80/$1 last; }`.

jacobtomlinson:
@betatim No it doesn't make a difference.

@manics Adding nginx or apache would probably work, but I would like to avoid that.

betatim:
We could add an option to (not) rewrite URLs to the proxy? `http://localhost:8888/lab/dont_rewrite_me_proxy/1234/` (maybe a slightly shorter URL ;) )

jacobtomlinson:
That sounds like a pragmatic fix. Perhaps something like `proxy_abs` for absolute urls?

psychemedia:
I've just come across this issue too using Apache Drill.

Includes are of the form:

```
<script src="/static/js/ace-code-editor/mode-snippets.js" type="text/javascript" charset="utf-8"></script>
```

Note also that the UI pages may themselves work at different path levels, eg `http://127.0.0.1:54239/proxy/8047/` or 
[...]

PR #84, opened 2019-01-11
Configure absolute and server URL prefixes

manics:
This started off as a fix for the `X-Forwarded-Context` `X-ProxyContextPath`. Previously they were always `/proxy/{port}`, now they should be `/server-name` if this is a managed server. Since this is related to the code that controls the URL rewriting I thought I'd add in support for absolute URLs.

I've done some testing with:
```
c.ServerProxy.servers = {
    'python-http': {
        'command': ['python3', '-m', 'http.server', '{port}'],
    },
    'python-http-abs': {
        'command': ['python3', '-m', 'http.server', '{port}'],
        'rewrite': '',
    },
}
```
`python-http` should produce logs like
```
"GET / HTTP/1.1" 200 -
```
`python-http-abs` should produce logs like
```
"GET /python-http-abs/ HTTP/1.1" 404 -
```
Also `/proxy-abs/port` is the absolute version of `/proxy/port`

Needs more testing! @jacobtomlinson @psychemedia 

Closes https://github.com/jupyterhub/jupyter-server-proxy/issues/43

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,11 @@ python:
 - 3.5
 
 script:
-  - true
+  - pip3 install .
+  - pip3 install pytest
+  - JUPYTER_TOKEN=secret jupyter-notebook --config=./tests/resources/jupyter_server_config.py &
+  - sleep 5
+  - pytest
 
 deploy:
   provider: pypi
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -4,17 +4,26 @@ Traitlets based configuration for jupyter_server_proxy
 from notebook.utils import url_path_join as ujoin
 from traitlets import Dict
 from traitlets.config import Configurable
-from .handlers import SuperviseAndProxyHandler, AddSlashHandler
+from .handlers import (
+    SuperviseAndProxyHandler,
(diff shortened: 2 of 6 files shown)


manics:
I've changed the config name from `absolute: False|True` to `rewrite: "/"|""` to match the terminology typically used when configuring web-servers and proxies, e.g. https://github.com/kubernetes/ingress-nginx/tree/nginx-0.21.0/docs/examples/rewrite

The prefix for ports without rewriting is still `/proxy-abs/port` though alternatives could be `/proxyp/{port}` (for proxy-passthrough) or `/passthrough/{port}`. What do you think?

yuvipanda:
w00t, this is awesome. I'll try take a closer look later today.

Vague first thoughts:

1. Are there ways we can do this without extra classes / inheritance? 
2. How about `/proxy/absolute`? That way we can do `/proxy/<options-string>` later if need be.

manics:
> Are there ways we can do this without extra classes / inheritance?

Yes, I think it's possible to combine the classes and use a more complex URL match, I'll give it a go

> How about /proxy/absolute

👍

manics:
Oops accidentally closed this .... reopened at https://github.com/jupyterhub/jupyter-server-proxy/pull/85

PR #85, opened 2019-01-16
Configure absolute and server URL prefixes

manics:
This started off as a fix for the `X-Forwarded-Context` `X-ProxyContextPath`. Previously they were always `/proxy/{port}`, now they should be `/server-name` if this is a managed server. Since this is related to the code that controls the URL rewriting I thought I'd add in support for absolute URLs.

I've done some testing with:
```
c.ServerProxy.servers = {
    'python-http': {
        'command': ['python3', '-m', 'http.server', '{port}'],
    },
    'python-http-abs': {
        'command': ['python3', '-m', 'http.server', '{port}'],
        'rewrite': '',
    },
}
```
`python-http` should produce logs like
```
"GET / HTTP/1.1" 200 -
```
`python-http-abs` should produce logs like
```
"GET /python-http-abs/ HTTP/1.1" 404 -
```
Also `/proxy-abs/port` is the absolute version of `/proxy/port`

Needs more testing! @jacobtomlinson @psychemedia 

Closes https://github.com/jupyterhub/jupyter-server-proxy/issues/43

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -138,6 +138,19 @@ class LocalProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             return url_path_join(self.base_url, self.proxy_base)
         return url_path_join(self.base_url, 'proxy', str(port))
 
+    def build_proxy_request(self, port, proxied_path, body, headers, options):
+        client_uri = '{uri}:{port}{path}'.format(
+            uri='http://localhost',
+            port=port,
+            path=proxied_path
+        )
+        if self.request.query:
+            client_uri += '?' + self.request.query
+
+        req = httpclient.HTTPRequest(
+            client_uri, method=self.request.method, body=body,
+            headers=headers, **options)
+        return req
 
     @web.authenticated
     async def proxy(self, port, proxied_path):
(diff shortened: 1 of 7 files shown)


manics:
Last two commits add the changes suggested by @yuvipanda (https://github.com/jupyterhub/jupyter-server-proxy/pull/84#issuecomment-454182484) and adds a travis test

yuvipanda:
@manics I've made the following changes here:

1. Call it `absolute_url` and make it boolean, instead of `rewrite`
2. Make it work with websockets too

This is an amazing contribution, thank you very much for it!

I'll merge shortly

yuvipanda:
@manics I'm actually going to wait for you to OK my changes before merging.

jupyter_server_proxy/handlers.py
```diff
@@ -142,32 +136,37 @@ def _get_context_path(self, port):
...
+
+    def _build_proxy_request(self, port, proxied_path, body):
+
         headers = self.proxy_request_headers()
 
+        client_uri = self.get_client_uri('http', port, proxied_path)
         # Some applications check X-Forwarded-Context and X-ProxyContextPath
         # headers to see if and where they are being proxied from.
-        if self.rewrite == '/':
+        if not self.absolute_url:
```
manics: @yuvipanda Since the `_build_proxy_request` method has been split you'll need either `context_path = self._get_context_path(port)` here, or have `self.get_client_uri` return a tuple `(client_uri, context_path)`

yuvipanda: Fixed!

yuvipanda: Fixed!

yuvipanda:
Am also super thankful for adding tests :) Would be nice to switch to Circle at some point though.

yuvipanda:
@manics I've made a new release (1.0.0b9) with your changes! \o/

Should really make a 1.0 release #86

Issue #87, opened 2019-01-19
Allow user to specify port


psychemedia:
From exploring running various services using `jupyter-server-proxy`:

- some services run on generally recognised ports, albeit allowing users to start pass a port number as an argument to the command that starts the service (eg OpenRefine on port 3333); downstream applications may assume this port number;
- some services appear to run on a default port that cannot be reassigned as part as the start-up command (eg pgAdmin4?)
- the parameter used to set the port may be arbitrarily named and as such hard to identify unambiguously in a command string;

It would be useful to be able to specify a default port number, e.g as:

```
c.ServerProxy.servers = {
    'aService': {
        'command': ['serviceCommand', '--p', '{port}'],
        'timeout': 120,
        'port' : MYPORT, #an optional argument specifying the required port number
        'launcher_entry': {
            'enabled': True,
            'title': 'myService',
        },
    },
}
```

See also: https://github.com/jupyterhub/jupyter-server-proxy/issues/81

ChandaniM123:
I am really looking forward for this change as I would like to pass the port the server process should run on.  Is this change coming anytime soon?

manics:
@psychemedia @ChandaniM123 Would you like to test https://github.com/jupyterhub/jupyter-server-proxy/pull/96? You can install it with `pip install https://github.com/manics/jupyter-server-proxy/archive/config-port.zip`

psychemedia:
@manics Seems to work fine on MyBinder.

(For some reason, `nbserverproxy` has stopped working under `repo2docker` locally on my Mac when it comes to trying to view UIs created on proxied ports.)

PR #96, opened 2019-01-31
Configure port for servers

manics:
Port can be specified for servers, e.g. to listen on port 54321 instead of a random port
```python
python-http-port54321': {
    'command': ['python3', './tests/resources/httpinfo.py', '{port}'],
    'port': 54321,
}
```
Closes https://github.com/jupyterhub/jupyter-server-proxy/issues/87

### DIFF
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -84,6 +84,12 @@ pairs.
    Defaults to *False*.
 
 
+ #. **port**
+
+     Set the port that the service will listen on. The default is to
+     automatically select an unused port.
+
+
 #. **launcher_entry**
 
    A dictionary with options on if / how an entry in the classic Jupyter Notebook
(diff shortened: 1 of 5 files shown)


yuvipanda:
@manics this is great! Can you add a doc for it in https://github.com/jupyterhub/jupyter-server-proxy/blob/master/docs/server-process.rst? can merge after

manics:
I've already added `**port**` under `Server Process options` https://github.com/jupyterhub/jupyter-server-proxy/pull/96/files#diff-f940d90ba22fa5e88b811022038fda33
What did you have in mind?

yuvipanda:
@manics I apparently need a new pair of eyeglasses...

yuvipanda:
<3 thank you for your work here! <3

Issue #86, opened 2019-01-18
Release 1.0


yuvipanda:
I think this is in a reasonable place to release a v1.0.

Primary tasks left to do are:

- [ ] Shore up documentation to be top notch - helping both users and developers understand what they can and can not do. 
- [ ] Make any final breaking API changes
- [ ] Document recommendation to use only in containers
- [ ] Move rsession-proxy into this repo
- [ ] Move docs of contrib packages into docs site

manics:
@yuvipanda I've just opened https://github.com/jupyterhub/jupyter-server-proxy/pull/96 once it's tested this repo will be in a good state for release

yuvipanda:
@manics I think you're right! The rest of the stuff can happen post 1.0 but pre-blog post.

yuvipanda:
I've now published 1.0! Thanks for the push, @manics!

I've opened #100 for merging rstudio back in here.

yuvipanda:
https://github.com/jupyterhub/jupyter-server-proxy/pull/101 for documentation on 'convenience' packages (new name for 'contrib' packages)

jakirkham:
Looks like it has been released. Should this be closed?

manics:
The doc improvements mentioned in the description are still on the to-do list.

PR #97, opened 2019-02-01
Bump version numbers to 1.0

yuvipanda:
Ref #86 

\o/

### DIFF
--- a/jupyterlab-server-proxy/package.json
+++ b/jupyterlab-server-proxy/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jupyterlab-server-proxy",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "description": "Launcher icons for proxied applications",
   "keywords": [
     "jupyter",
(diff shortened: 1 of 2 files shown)


yuvipanda:
/cc @manics and @ryanlovett! 

I still wanna do the doc bits in #86 before we publicize this very well though.

If you give me your PyPI and npm usernames I can add you as maintainers there.

ryanlovett:
> If you give me your PyPI and npm usernames I can add you as maintainers there.

rylo on PyPI, ryanlovett on npmjs.

yuvipanda:
I think there's automation in this repo that pushes to PyPI automatically on tags. Would that work for the new name (jupyter-server-proxy) too? We have to be careful to not break the older nbserverproxy package.

ryanlovett:
If pypi distinguishes between nbserverproxy and jupyter-server-proxy without one being handled as an alias for the other, then in the short term we should disable the travis deploy and then later update it for jupyter-server-proxy or setup setup a new CI with Circle.

Currently we deploy through an "nbserverproxy" pypi user which has rights to the "nbserverproxy" pypi project.

manics:
@yuvipanda Sorry for the delay, my pypi username is `sas`, npm username (just created!) is `manics`

PR #99, opened 2019-02-04
Add release docs & remove automated publish to PyPI

yuvipanda:
We can add it back later if we wish - especially once we
get more packages into `contrib/`


yuvipanda:
@ryanlovett I removed the automation and added a release checklist instead. What do you think?

PR #100, opened 2019-02-04
Bring in https://github.com/jupyterhub/jupyter-rsession-proxy

yuvipanda:
In an attempt to have just one big repository with multiple
helper packages for each popular package we might wanna run,
this merges RStudio support under 'contrib'. We wanna keep the
history to make sure we credit everyone involved in making
this possible.


ryanlovett:
> who did almost all the work here.

Sort contributions by additions/deletions, not commits :) . Thanks @yuvipanda for elevating it to something real (stability, usability, etc.) and not just a proof of concept.

Issue #112, opened 2019-03-17
Configure proxy request options with traitlets


ryanlovett:
One can subclass to specify the proxy request options, but it'd be convenient not to have to do that. nbstencila alters request timeouts and I've found I'll need to do the same for syncthing.

PR #113, opened 2019-03-18
Move proxy_request_options to configuration.

ryanlovett:
To address #112.

### DIFF
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -107,6 +107,13 @@ pairs.
    #. **title**
       Title to be used for the launcher entry. Defaults to the name of the server if missing.
 
+
+#. **proxy_request_options**
+
+   A dictionary of options to be used when constructing a
+   tornado.httpclient.HTTPRequest instance for the proxy request.
+   The default is ``{ follow_redirects=False }``.
+
 .. _server-processes/callable-arguments:
 
 Callable arguments
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -9,7 +9,7 @@ import pkg_resources
 from collections import namedtuple
 from .utils import call_with_asked_args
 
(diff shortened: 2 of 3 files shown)


ryanlovett:
Example app that requires a proxy request option:
https://github.com/ryanlovett/jupyter-syncthing

yuvipanda:
This is awesome, and we should definitely support configuring timeouts. However, I think we should explicitly support them, since otherwise we will be tied down to Tornado's HTTP Client implementation. I'd love to move to aiohttp at some point.

costa:
Hello, thank you for your effort.

I need to configure jupyterhub proxy timeouts (without monkey-patching the code preferably).

I'll just wait for this to be merged, right?

It cannot take long, definitely under half a year, right?

manics:
@costa Are you able to help test this?

costa:
@manics Let's see, I'm using the JupyterHub helm chart with a custom-built singleuser image which is built with installing `jupyter-server-proxy` through pip; I guess I can install a custom version of that to produce a test image to use with a test JupyterHub deployment — if it is not too much effort to do so.

costa:
Guys (no assumption here ;)), this PR has conflicts, I'm probably willing to test this if I have a chance to change the proxy request timeout without having to monkey-patch the code myself — and anyone can (fix the conflicts and) confirm it's worth the effort.

yuvipanda:
> It cannot take long, definitely under half a year, right?

:D 

I think unfortunately this PR didn't make it.

While we do have a process ready timeout (https://github.com/jupyterhub/jupyter-server-proxy/blob/75cba2cef9765d56aa5374bb6de7d4d89a59cbf3/docs/source/server-process.md?plain=1#L41) we don't have a request timeout.

I think we should implement request timeout as its own option, rather than allowing arbitrary options to be passed through. That exposes internal detail of the project (which http client it uses) outside, making it impossible to change.

With that, I'm going to close this.

Issue #39, opened 2018-06-17
502 During Web Socket Handshake


rmorshea:
I'm trying to get web sockets to work when running a [Sanic](http://sanic.readthedocs.io/en/latest/sanic/routing.html#websocket-routes) server via proxy on https://mybinder.org.

I don't think I know enough to understand whether or not the 502 Bad Gateway is resulting from a problem with Sanic or with nbserverproxy (or the way I've configured either). Is this a common issue? If not, what information do I need in order to debug this?

@ryanlovett I noticed that you worked on the socket proxy logic - any thoughts?

ryanlovett:
Hi @rmorshea , do you have an example Sanic server? I haven't received other reports of Sanic servers, good or bad.

I tried https://github.com/Enforcer/simple-chat with the patch below and it worked for me on a JupyterHub instance. I'll try on binder.
```
--- static/index.html.orig	2018-06-18 17:11:07.000000000 -0700
+++ static/index.html	2018-06-18 17:10:58.000000000 -0700
@@ -63,7 +63,8 @@


 $(function() {
-    websocket = new WebSocket('ws://127.0.0.1:8000/chat');
+    var url = 'wss://' + document.location.hostname + document.location.pathname + '/chat';
+    websocket = new WebSocket(url);
     websocket.onopen = init(websocket);
     websocket.onerror = websocket.onclose = close;
     websocket.onmessage = newMessage;
```

ryanlovett:
Works on binder too, https://mybinder.org/v2/gh/ryanlovett/simple-chat/master. Start server.py in a terminal and visit proxy/8000.

That being said, maybe nbserverproxy isn't doing the right thing for more complicated ws apps.

joshbode:
I think I'm seeing something similar with kubernetes + jupyterhub + jupyterlab + nbserverproxy + dask:

From the jupyterlab logs:
```
[I 2018-07-30 02:05:10.484 SingleUserLabApp log:158] 200 GET /user/josh.bode/proxy/35902/status (josh.bode@10.142.0.6) 123.54ms
[I 2018-07-30 02:05:12.053 SingleUserLabApp handlers:299] Client sent subprotocols: []
Future exception was never retrieved
future: <Future finished exception=IndexError('list index out of range',)>
Traceback (most recent call last):
  File "/opt/conda/lib/python3.6/site-packages/tornado/gen.py", line 326, in wrapper
    yielded = next(result)
[...]

PR #119, opened 2019-04-06
WebsocketHandler.get is async

rmorshea:
As of Tornado v6.0.2 some methods were converted to be async/await capable. See commit:

https://github.com/tornadoweb/tornado/commit/e69becbf8c994be5bfa6584ef72df594b27934b5

Closes: https://github.com/jupyterhub/jupyter-server-proxy/issues/39

### DIFF
--- a/jupyter_server_proxy/websocket.py
+++ b/jupyter_server_proxy/websocket.py
@@ -93,7 +93,7 @@ class WebSocketHandlerMixin(websocket.WebSocketHandler):
         if self.request.headers.get("Upgrade", "").lower() != 'websocket':
             return await self.http_get(*args, **kwargs)
         else:
-            awwait maybe_future(super().get(*args, **kwargs))
+            await maybe_future(super().get(*args, **kwargs))
 
 
 def setup_handlers(web_app):
--- a/jupyter_server_proxy/websocket.py
+++ b/jupyter_server_proxy/websocket.py
@@ -85,8 +85,7 @@ class WebSocketHandlerMixin(websocket.WebSocketHandler):
     async def get(self, *args, **kwargs):
         if self.request.headers.get("Upgrade", "").lower() != 'websocket':
             return await self.http_get(*args, **kwargs)
-        # super get is not async
-        super().get(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


betatim:
How could we deal with being compatible with Tornado 5 and 6 at the same time?

I was thinking https://www.tornadoweb.org/en/stable/gen.html#tornado.gen.maybe_future but that is deprecated. So maybe https://docs.python.org/3/library/inspect.html#inspect.isawaitable is the right thing to use if we should `await` or not?

rmorshea:
Got it. Will make that change tomorrow.

rmorshea:
@betatim done!

ian-r-rose:
Recently the notebook server swapped out tornado's `maybe_future` for its own to deal with that deprecation: https://github.com/jupyter/notebook/pull/4453 . That may be a good choice here as well.

rmorshea:
@ian-r-rose does this problem show up in other places in this repo? I think it makes sense to use that if there's more locations where we need to address the backward compatibility.

ian-r-rose:
@rmorshea I'm not sure. This seems to fix the immediate problem, but there may be more lurking somewhere. I confess I don't have the best grasp of websockets.

rmorshea:
@ian-r-rose neither do I. @betatim actually debugged #39 I just made the PR.

betatim:
I don't really know either. Having looked at the [implementation of `maybe_future` in the notebook](https://github.com/jupyter/notebook/pull/4453) I am now thinking that we should use that. As with all these things it isn't as easy as just asking `if isawaitable(obj)`. Especially as it seems that in this thread it is mostly the blind leading the blind 👀.

Does this mean that jupyter-server-proxy will require a brand new notebook package in order to work (if we use `maybe_future` from there)? If so that is also not ideal.

ian-r-rose:
@betatim That's a good point. Perhaps we should try to import `maybe_future`, then fall back on `gen.maybe_future` if that fails?

rmorshea:
@betatim used a `try/except` to import from `notebook` and then from `tornado`.

betatim:
Merged. Thanks for this!

Issue #70, opened 2019-01-02
Usage with `dask-labextension`


ian-r-rose:
This is a follow-up with from some gitter chatting with @yuvipanda.

The [dask-labextension](https://github.com/dask/dask-labextension) JupyterLab extension currently involves adding iframe'd bokeh dashboard panels to the main work area. In order to get around CORS issues, it includes a [modified version](https://github.com/dask/dask-labextension/blob/master/dask_labextension/dashboardhandler.py) of `nbserverproxy` to proxy them under the notebook server origin. It is possible to do the same thing with an unmodified version of the package, but there were a couple of things that made me include the modified version:

1. I would like to leave open the possibility that the dashboard URL is not on `localhost`. Our current examples are all on `localhost`, but I don't view that as a requirement that needs to be enforced, and we want to be able to support a wide variety of dask clusters which may have different deployment patterns that we haven't necessarily thought of.
2. The dask labextension includes rest endpoints for cluster management as well as dashboard proxying. Currently, the cluster management endpoint is at `/dask/clusters/<cluster-id>`, and the corresponding dashboard is at `/dask/dashboards/<cluster-id>`. I'd like to be able to continue this pattern.

So there are two things that would be nice for `nbserverproxy` (or `jupyter-server-proxy`) to allow to be configured:
1. The ability to proxy to arbitrary URLs, rather than just ports on `localhost`.
2. The ability to place a given proxy at a different endpoint on the notebook server than `/proxy/<port>/<path>`.

I know that @yuvipanda has put some significant work into revamping this package recently, so it is possible that one can already do this. If that's the case, do let me know!

ryanlovett:
@jacobtomlinson has a patch to address the first item on your wish list at https://github.com/jupyterhub/nbserverproxy/pull/46.

You can already do the second item. See how [nbstencila accomplishes it](https://github.com/minrk/nbstencilaproxy/blob/master/nbstencilaproxy/handlers.py#L64) in the latest release and how [nbrsessionproxy does it](https://github.com/jupyterhub/nbrsessionproxy/blob/master/setup.py#L18) in master.

ian-r-rose:
Thanks for the info @ryanlovett! I'll do some experiments with the current overhauled version and see how far I can get.

yuvipanda:
@ian-r-rose hold on for a day or two, and I think it'll be clearer.

ian-r-rose:
[...]

PR #128, opened 2019-05-02
Allow for other hosts than `localhost` to be proxied.

ian-r-rose:
Finishes the work needed for #70, dask/dask-labextension#58.

This allows the developer to specify a non-localhost URI to be proxied. It should be a non-breaking change for users to `LocalProxyHandler`. It seems to work well in my testing, but @yuvipanda and @ryanlovett have a much better idea of the range of places where `jupyter-server-proxy` are deployed.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -31,59 +31,15 @@ class AddSlashHandler(IPythonHandler):
 class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
     """
     A tornado request handler that proxies HTTP and websockets from
-    a given host/port combination.
+    a given host/port combination. This class is not meant to be
+    used directly as a means of overriding CORS. This presents significant
+    security risks, and could allow arbitrary remote code access. Instead, it is
+    meant to be subclassed and used for proxying URLs from trusted sources.
     """
     def __init__(self, *args, **kwargs):
         self.proxy_base = ''
         self.absolute_url = kwargs.pop('absolute_url', False)
         super().__init__(*args, **kwargs)
-
-    async def open(self, host, port, proxied_path=''):
-        """
-        Called when a client opens a websocket connection.
-
-        We establish a websocket connection to the proxied backend &
-        set up a callback to relay messages through.
-        """
-        if not proxied_path.startswith('/'):
-            proxied_path = '/' + proxied_path
-
-        client_uri = self.get_client_uri('ws', host, port, proxied_path)
-        headers = self.request.headers
-
-        def message_cb(message):
-            """
-            Callback when the backend sends messages to us
-
-            We just pass it back to the frontend
-            """
-            # Websockets support both string (utf-8) and binary data, so let's
-            # make sure we signal that appropriately when proxying
-            self._record_activity()
-            if message is None:
-                self.close()
-            else:
-                self.write_message(message, binary=isinstance(message, bytes))
-
-        def ping_cb(data):
-            """
-            Callback when the backend sends pings to us.
-
-            We just pass it back to the frontend.
-            """
-            self._record_activity()
-            self.ping(data)
-
-        async def start_websocket_connection():
-            self.log.info('Trying to establish websocket connection to {}'.format(client_uri))
-            self._record_activity()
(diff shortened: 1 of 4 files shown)


jupyter_server_proxy/handlers.py
```diff
@@ -28,14 +28,17 @@ def get(self, *args):
         dest = src._replace(path=src.path + '/')
         self.redirect(urlunparse(dest))
 
-class LocalProxyHandler(WebSocketHandlerMixin, IPythonHandler):
-
+class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
```
yuvipanda: Can you make a note here about how this should be only used by classes that are overriding this from someplace else, and not directly, to prevent bypassing CORS?

jupyter_server_proxy/handlers.py
```diff
@@ -242,27 +245,27 @@ def proxy_request_options(self):
     # Support all the methods that torando does by default except for GET which
     # is passed to WebSocketHandlerMixin and then to WebSocketHandler.
 
-    async def http_get(self, port, proxy_path=''):
+    async def http_get(self, host, port, proxy_path=''):
```
yuvipanda: I *think* we should remove all these definitions - this handler should only be used by inherited handlers, and these need to be manually defined there anyway. And where they are defined manually, you must not let the user define which host it is...

jupyter_server_proxy/handlers.py
```diff
@@ -28,14 +28,17 @@ def get(self, *args):
...
+    A tornado request handler that proxies HTTP and websockets from
+    a given host/port combination.
+    """
     def __init__(self, *args, **kwargs):
         self.proxy_base = ''
         self.absolute_url = kwargs.pop('absolute_url', False)
         super().__init__(*args, **kwargs)
 
-    async def open(self, port, proxied_path=''):
+    async def open(self, host, port, proxied_path=''):
```
yuvipanda: I somehow don't want this to be 'open' here, since that too can be accidentally used directly by users to gain access to arbitrary websocket domains without CORS protection. Can we move this open method to be called proxy_open (to match the 'proxy' method), and call it from 'open' in LocalProxyHandler and friends? This makes it so that each handler that inherits from this *must* implement its own methods for open, http_get, put, post, etc.

jupyter_server_proxy/handlers.py
ian-r-rose: Interesting idea. I'm a little reticent to do something to break this class in such an implicit way -- that will make developing and debugging the subclassed handler trickier when there is no indication of what methods to override.

What if instead we make all these functions raise an error with a useful message?

jupyter_server_proxy/handlers.py
ian-r-rose: Sure.

jupyter_server_proxy/handlers.py
ian-r-rose: Makes sense.

jupyter_server_proxy/handlers.py
ian-r-rose: Okay, I've taken a crack at what I suggested. Let me know what you think.

yuvipanda:
@ian-r-rose this looks good to me! Unfortunately there are no tests :| Can you do a bunch of local testing to see if it works ok? Would be even better if you can add a quick test or two, but I understand if you don't want this to block on that.

ian-r-rose:
I've tested the `LocalProxyHandler` with dask-labextension, as well as ian-r-rose/jupyterlab-bokeh-server, and both of those work without changes. I've also tested `dask-labextension` with `ProxyHandler` using `127.0.0.1` for the hostname.

I can try to add some tests this evening.

ian-r-rose:
I assume you also have some of extensions which it might be good to check with @yuvipanda?

yuvipanda:
@ian-r-rose would <3 if you could add some simple tests :)

ian-r-rose:
I've added a very basic test, the great majority of the effort for which went into setting up the serverextension.

yuvipanda:
Thanks for all the awesome work, @ian-r-rose! <3

Issue #62, opened 2018-12-21
Support for websocket compression


snarisi:
This is currently failing for me when proxying from a websocket client that requests permessage-deflate compression:

```
    Traceback (most recent call last):
      File "/opt/conda/lib/python3.6/site-packages/tornado/http1connection.py", line 185, in _read_message
        header_future = delegate.headers_received(start_line, headers)
      File "/opt/conda/lib/python3.6/site-packages/tornado/http1connection.py", line 651, in headers_received
        return self._delegate.headers_received(start_line, headers)
      File "/opt/conda/lib/python3.6/site-packages/tornado/websocket.py", line 1185, in headers_received
        self.protocol._process_server_headers(self.key, self.headers)
      File "/opt/conda/lib/python3.6/site-packages/tornado/websocket.py", line 783, in _process_server_headers
        raise ValueError("unsupported extension %r", ext)
    ValueError: ('unsupported extension %r', ('permessage-deflate', {'client_max_window_bits': '15'}))
```

I was able to get in working in my case by changing the source code to instantiate the `PingableWSClientConnection` with `compression_options={}`. Not sure if this is the correct fix, or if there's a better way to add support for compression.

PR #136, opened 2019-07-04
Support websocket compression

rcolle:
`jupyter-server-proxy`'s websocket connection (`PingableWSClientConnection`) does not support WS compression.

When using `jupyter-server-proxy` to proxy a websocket connection, the handshake that occurs during the upgrade only takes into account the proxied websocket server's preferences. That server may (and often will) support compression schemes such as `permessage-deflate`.
In this case, the client will use that compression scheme which will trigger an error in `jupyter-server-proxy` / `tornado` code.
This is exactly what happens in #62 .

The fix is fairly simple: `PingableWSClientConnection` should support all compression schemes and let the client and the proxied server decide how they should communicate.


yuvipanda:
Thanks for making the PR, @rcolle!!!

PR #140, opened 2019-07-24
Fix WebSocket race condition

tibdex:
Before this fix, we could go through https://github.com/jupyterhub/jupyter-server-proxy/blob/90401f1290bc3c5b9408392f8f3b93bde5f55919/jupyter_server_proxy/handlers.py#L82-L83 before the future here https://github.com/jupyterhub/jupyter-server-proxy/blob/90401f1290bc3c5b9408392f8f3b93bde5f55919/jupyter_server_proxy/handlers.py#L262-L263 resolves.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -231,6 +231,8 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
 
         client_uri = self.get_client_uri('ws', host, port, proxied_path)
         headers = self.request.headers
+        current_loop = ioloop.IOLoop.current()
+        ws_connected = current_loop.asyncio_loop.create_future()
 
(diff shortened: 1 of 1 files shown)


tibdex:
Hello, it's been almost a month without replies. Is there something I can do to help get this fixed and published on PyPI?

manics:
The change sounds reasonable to me, but I don't have enough knowledge of async/await to review it. Hopefully someone else can.

betatim:
Thanks!

betatim:
To get a release quickly some work on extending the travis.yml would help so that releases happen automatically when we tag a release. Right now someone has to do the releasing manually which means it doesn't get done very often :-/

ryanlovett:
@betatim Could you please file an issue for that? :)

PR #148, opened 2019-09-22
Add code-server-traitlet example.

ryanlovett:
Traitlet derivative of @betatim's vscode-binder.

### DIFF
--- a/contrib/code-server-traitlet/README.md
+++ b/contrib/code-server-traitlet/README.md
@@ -1,0 +1,5 @@
+## code-server-traitlet
+
+Create a code-server launcher [via traitlet](https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#specifying-config-via-traitlets).
+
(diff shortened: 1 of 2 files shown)


betatim:
Nice! How would I use this in a repo like (say) https://github.com/binder-examples/requirements? Basically a repo where I want to have VS Code as editor available but unlike the vscode-binder repo I don't want all that cruft/extra code?

ryanlovett:
You can drop the jupyter_notebook_config.py into a jupyter config directory, e.g. ~/.jupyter/, replacing your jupyter_vscode_proxy extension.

Here's an [example](https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#specifying-config-via-traitlets) which demonstrates the same with openrefine which you also had made a package for way back when. ;)

betatim:
Ah ok. So I'd add that config file, the executables needed and be done!

Merging.

PR #151, opened 2019-09-26
Optionally map requested → proxied paths

manics:
Most webservers can be configured to serve a page if a directory `/` is accessed, or to redirect to a different path. This PR adds this functionality to jupyter-proxy-server.

For example, when using noVNC the default page is a directory listing. With this PR you can map `/` to `vnc_lite.html`.

E.g. https://gist.github.com/manics/7d1f4b76ce06c2bb07db88e3496a1561

### DIFF
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -90,6 +90,13 @@ pairs.
      automatically select an unused port.
 
 
+#. **mappath**
+
+     Map request paths to proxied paths.
+     Either a dictionary of request paths to proxied paths,
+     or a callable that takes parameter ``path`` and returns the proxied path.
+
+
 #. **launcher_entry**
 
    A dictionary with options on if / how an entry in the classic Jupyter Notebook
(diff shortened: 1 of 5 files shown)


ryanlovett:
@manics Nice! I think this is more akin to redirecting a specific route rather than indexing, e.g. nginx's proxy_redirect. It'd be cool if it could be generalized somehow, but I'm guessing the `/` case is common.

manics:
@ryanlovett Good idea! I've generalised it so the parameter is called `mappath`, and can be a dictionary or a callable.

Also updated the example https://gist.github.com/manics/7d1f4b76ce06c2bb07db88e3496a1561

ryanlovett:
Awesome, thanks!

prateekkhera:
@ryanlovett - sorry, seems the page moved, can you paste the example here please for `mappath`. I'll be grateful, thanks!

PR #154, opened 2019-10-01
Allow proxying to remote host

### LINKED ISSUE #152: Proxy to host other than localhost
It would be really helpful to be able to proxy to another service which is only accessible from the notebook.

My use case here is a Dask cluster running on a Kubernetes cluster that is also running the notebook (z2hj + Pangeo). The notebook can connect to the scheduler at its hostname, but I cannot view the dashboard because the hostname and IP of the dashboard is only routable from within the Kubernetes cluster.

It would be great to be able to do something like:

`/proxy/{hostname}:{port}/`

_This has come up for the first time as [work in `dask-kubernetes`](https://github.com/dask/dask-kubernetes/pull/162) has meant that the scheduler will no longer be running in on the same host as the notebook, and instead be a separate container within the cluster with its own IP and hostname._

jacobtomlinson:
This PR fixes #152 by allowing you to proxy to another host using the format:

```
/proxy/{host}:{port}/{path}
```

I appreciate that there may be some security concerns around this but considering that Jupyter already provides a remote execution environment and a bad actor could simply use `nc` along with the existing proxy to acheive port forwarding to any host within the network I don't think this increases the existing attack surface.

Feedback and discussion very welcome!


ian-r-rose:
I can take a look at this later, but for the moment there is some discussion of the security implications of this in #128.

yuvipanda:
Thanks for opening this, @jacobtomlinson!

I'm mostly worried about bypassing cross origin protection in *other* third party websites through this from malicious js that can't otherwise run backend code.

I'd like us to whitelist domains with a configurable here to whitelist domains. How does that sound?

jacobtomlinson:
Ah I see. 

My only concern is that the dask-kubernetes service names are random UUIDs and so will be unique for every cluster. Could there be a way to dynamically whitelist? A fixed whitelist wouldn't work for this use case.

manics:
How about allowing a regex or glob in the whitelist?

jacobtomlinson:
That would work, are you happy with that @yuvipanda?

jacobtomlinson:
I've quickly implemented this using a regex based whitelist. I'm not familiar enough with the project to have a feeling for the best way to get the config from within the handler.

https://github.com/jupyterhub/jupyter-server-proxy/pull/154/files#diff-bc4411a876f25fa78f564f274f779fd9R171

Coud someone please point me in the right direction here?

mrocklin:
Checking in, what is the current status on this?  @jacobtomlinson is this blocking the dask-kubernetes rewrite and release?

manics:
I've just looked through some of the related issues, I'm a bit confused as to how in your use case (internally hosted dask dashboard) a user will know the `/proxy/{host}:{port}/{path}` to go to since it's an external process, not managed by jupyter-server-proxy's simpervisor.

For config I'd guess @yuvipanda meant adding a new top-level config property to https://github.com/jupyterhub/jupyter-server-proxy/blob/0e67e1afd0bab1342443f13bd147a2f8c682e9e0/jupyter_server_proxy/config.py#L117

jacobtomlinson:
@manics Dask knows the address of the dashboard. This is provided programmatically to the Dask JL Extension and also the widget (with a link the user clicks). 

Thanks for the pointer to the config!

@mrocklin yeah this is blocking the release.

yuvipanda:
When me and @ian-r-rose had this conversation last, we ended up with https://github.com/dask/dask-labextension/blob/master/dask_labextension/dashboardhandler.py#L16. You can proxy to remote hosts by overriding this in your own extension. Does your rewrite have your own extension to put that code in? If so, IMO that's the 'right' thing to do - you control the hosts, but you also control the URL path to be whatever you want. If it's in config, you still have to find a way to add it to the jupyter_notebook_config.py file.

If you *do* want this to be a config, I think having that be function that can dynamically take a hostname and return True / False is better than a RegEx.

Thank you for working on this! I'm sorry about the tardy responses. Happy to spend some time in a more immediate medium (chat / voice) if that'll help.

yuvipanda:
As for creating config, you'll have to use [traitlets](https://traitlets.readthedocs.io/en/stable/)

nbresuse offers useful example:

1. https://github.com/yuvipanda/nbresuse/blob/master/nbresuse/__init__.py#L56 creates a Config object, and defines the aspects that can be configurable
2. https://github.com/yuvipanda/nbresuse/blob/master/nbresuse/__init__.py#L96 it's populated from jupyter_notebook_config.py (and other files Jupyter checks) and put in a place that web handlers can see
3. Accessed in the handlers https://github.com/yuvipanda/nbresuse/blob/master/nbresuse/__init__.py#L17

jacobtomlinson:
@yuvipanda This isn't for an extension so I don't think I have anywhere to write a custom handler. 

Not used traitlets yet, looks neat. I'll have a go at adding it, but if I struggle I may ping you for a sync.

rcthomas:
@jacobtomlinson is the only thing needed here the traitlets config stuff for the whitelist function?

jacobtomlinson:
Yes but this has dropped down my priority list slightly.

rcthomas:
@jacobtomlinson if you review+merge the PR I just sent you then I think that might address the configuration question and then @yuvipanda can review.

jacobtomlinson:
@rcthomas done, many thanks for pushing this along.

rcthomas:
There might be another thing to fix here I realized.  The handlers need to be swapped (make Remote handler first) or the regex match in the handlers refined more.  When I was testing my fix I was using a setup that had a Docker network and a name like "service1:8000" and that matched the proper handler.  But ordering matters, and when I tried late last night with "10.128.0.20:8787" it matched the Local one (because \d's matched).  I did a live fix swapping the handlers and that made the behavior correct.

jupyter_server_proxy/handlers.py
```diff
@@ -467,13 +521,17 @@ def options(self, path):
...
     host_pattern = '.*$'
     web_app.add_handlers('.*', [
         (url_path_join(web_app.settings['base_url'], r'/proxy/(\d+)(.*)'),
          LocalProxyHandler, {'absolute_url': False}),
         (url_path_join(web_app.settings['base_url'], r'/proxy/absolute/(\d+)(.*)'),
          LocalProxyHandler, {'absolute_url': True}),
+        (url_path_join(web_app.settings['base_url'], r'/proxy/(.*):(\d+)(.*)'),
+         RemoteProxyHandler, {'absolute_url': False, 'host_whitelist_hook': host_whitelist_hook}),
+        (url_path_join(web_app.settings['base_url'], r'/proxy/absolute/(.*):(\d+)(.*)'),
+         RemoteProxyHandler, {'absolute_url': True, 'host_whitelist_hook': host_whitelist_hook}),
```
jacobtomlinson: Do you mean the order here needs changing?

rcthomas: Yes, the `RemoteProxyHandler`s need to be added before the `LocalProxyHandler` ones, since Tornado stops at the first match.  If it's an IP, then the `LocalProxyHandler`s match up to the end of the first quad and that one will fire.  More importantly, I think the presence of the ":" is the thing that should make or break Remote vs Local.

jacobtomlinson: @rcthomas done!

rcthomas:
I think that looks good.  @yuvipanda What do you think?

jupyter_server_proxy/handlers.py
```diff
@@ -10,6 +10,7 @@
 from urllib.parse import urlunparse, urlparse
 import aiohttp
 from asyncio import Lock
+import re
```
manics:
```suggestion
```

jupyter_server_proxy/config.py
```diff
@@ -162,3 +162,25 @@ class ServerProxy(Configurable):
         """,
         config=True
     )
+
+    host_whitelist_hook = Any(
```
yuvipanda: I think this should instead be a Union of a List or Callable, and be called 'host_whitelist'. So you can either specify a list (which can then also be passed in via the commandline on starting notebook), or a callable.

There's an example of how to do that in https://github.com/yuvipanda/nbresuse/blob/master/nbresuse/__init__.py#L136.

rcthomas: Agreed.

rcthomas: An issue is traitlets 4.3.3 doesn't include `Callable()` and it seems to be the most recent release thought it's there in master.

yuvipanda: @rcthomas yeah, unfortunately you need to copy the implementation. That's what nbresuse does

jacobtomlinson: This should be in now thanks to @rcthomas

yuvipanda:
Thanks, @rcthomas and @jacobtomlinson! I think this is good to go once the merge conflict is resolved.

rcthomas:
I think it's just here https://github.com/jupyterhub/jupyter-server-proxy/blob/master/jupyter_server_proxy/config.py#L12 @jacobtomlinson do you want to handle that?  I probably should have just submitted a different PR directly, but we're almost done here...

yuvipanda:
w00t, thanks @rcthomas and @jacobtomlinson!

I'll try make a release shortly

mrocklin:
Thanks everyone!

jacobtomlinson:
Thanks all!

yuvipanda:
Released!

PR #159, opened 2019-11-06
Move npm package to jupyterlab org

blink1073:
This allows JupyterLab maintainers to make the npm releases.


blink1073:
Thanks @ian-r-rose!  Published and public: https://www.npmjs.com/package/@jupyterlab/server-proxy

Issue #162, opened 2019-11-27
icon_url should be replaced with icon_path in the cookiecutter template


ka-zo:
Hi @yuvipanda ,

There is an error in the following line:

https://github.com/jupyterhub/jupyter-server-proxy/blob/a937a03fcc9bb21d9cf6b39eefba68915187e0dd/contrib/template/%7B%7Bcookiecutter.project_name%7D%7D/jupyter_%7B%7Bcookiecutter.project_name%7D%7D_proxy/__init__.py#L15

`icon_url` should be `icon_path`. Please fix, otherwise icon shall not be loaded for the GUI launcher, after cookiecutter generates the project from the template.

betatim:
As you've found the problem and diagnosed it, do you want to make a PR to fix this?

ka-zo:
ok. no problem.

PR #163, opened 2019-11-28
Replace icon_url with icon_path in __init__.py.

ka-zo:
In line with the documentation the correct key for the path of the icons folder in the launcher_entry is icon_path and not icon_url. See https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html

This PR is a solution for issue #162


betatim:
Thanks a lot!

PR #174, opened 2020-03-04
Support usage of JupyterLab 2.0.0

### LINKED ISSUE #173: JupyterLab 2.0.0 compatibility
JupyterLab 2.0.0 is out, and this extension have conflicting dependencies with its core packages.

```
Conflicting Dependencies:
JupyterLab              Extension      Package
>=2.0.0 <2.1.0          >=1.0.0 <2.0.0 @jupyterlab/application
>=2.0.0 <2.1.0          >=1.0.0 <2.0.0 @jupyterlab/launcher
```

## References
- [Change log for developers](https://jupyterlab.readthedocs.io/en/stable/getting_started/changelog.html#for-developers)
- [A migration guide](https://jupyterlab.readthedocs.io/en/stable/developer/extension_migration.html#extension-migration)

consideRatio:
Fixes #173

At least, I think this fixes #173, I have not verified the functionality but only verified I could build it without failures after bumping it. Also, note the comment in the second commit for an explanation about the change in tsconfig.json

### DIFF
--- a/jupyterlab-server-proxy/tsconfig.json
+++ b/jupyterlab-server-proxy/tsconfig.json
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
     "declaration": true,
     "lib": ["es2015", "dom"],
     "module": "commonjs",
--- a/jupyterlab-server-proxy/package.json
+++ b/jupyterlab-server-proxy/package.json
@@ -30,12 +30,12 @@
     "watch": "tsc -w"
   },
   "dependencies": {
-    "@jupyterlab/application": "^1.0.0",
-    "@jupyterlab/launcher": "^1.0.0"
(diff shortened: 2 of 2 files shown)


betatim:
Do you think we will be able to use this with JupyterLab 1.0 and 2.0 simultaneously or we need separate builds/releases?

consideRatio:
I think simultaneous use is fine as I didn't change the code at all.

jupyterlab-server-proxy/tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
```
betatim: On purpose or a accidental change?

consideRatio: Om purpose, the reason is described in the second commit's commit messages

PR #177, opened 2020-03-14
Quote proxied_path to make it a safe URL (fixes #138)

### LINKED ISSUE #138: URLs with pathname containing spaces not supported
Hi,

Trying to request an URL with a [pathname](https://nodejs.org/api/url.html#url_url_pathname) containing spaces through the proxy leads to HTTP errors such as `400 (Bad request)` or `505 (HTTP Version Not Supported)`.

Here's how you can reproduce:

1. Launch this Node.js server:
   ```javascript
   const http = require("http");
   const port = 3000;

   const requestHandler = (request, response) => {
     response.end(request.url);
   };

   const server = http.createServer(requestHandler);

   server.listen(port, error => {
     if (error) {
       return console.error(error);
     }

     console.log(`listening on ${port}`);
   });
   ```
2. Start a Jupyter server with `jupyter-server-proxy` active
3. Check in your browser that http://localhost:3000/foo%20bar works but http://localhost:8888/proxy/3000/foo%20bar doesn't

Thanks!

oeway:
Currently if there is special character in the proxied path, it failed with error, e.g:
```bash
[E 16:28:29.202 NotebookApp] Uncaught exception GET /myservice/static/Desktop/Screenshot%202020-03-10%20at%2021.50.46.png (127.0.0.1)
    HTTPServerRequest(protocol='http', host='127.0.0.1:9527', method='GET', uri='/myservice/static/Desktop/Screenshot%202020-03-10%20at%2021.50.46.png', version='HTTP/1.1', remote_ip='127.0.0.1')
      File "/Users/me/miniconda3/envs/myservice/lib/python3.6/site-packages/tornado/web.py", line 1703, in _execute
        result = await result
      File "/Users/me/miniconda3/envs/myservice/lib/python3.6/site-packages/jupyter_server_proxy/websocket.py", line 96, in get
        return await self.http_get(*args, **kwargs)
      File "/Users/me/miniconda3/envs/myservice/lib/python3.6/site-packages/jupyter_server_proxy/handlers.py", line 510, in http_get
        return await self.proxy(self.port, path)
      File "/Users/me/miniconda3/envs/myservice/lib/python3.6/site-packages/jupyter_server_proxy/handlers.py", line 503, in proxy
        return await super().proxy(self.port, path)
      File "/Users/me/miniconda3/envs/myservice/lib/python3.6/site-packages/jupyter_server_proxy/handlers.py", line 213, in proxy
        response = await client.fetch(req, raise_error=False)
    tornado.simple_httpclient.HTTPStreamClosedError: Stream closed
```

This is a fix for that by quote the proxied path and make it safe.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -140,7 +140,7 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
         else:
             client_path = proxied_path
 
-        client_path = quote(client_path)
+        # client_path = quote(client_path)
 
         client_uri = '{protocol}://{host}:{port}{path}'.format(
             protocol=protocol,
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -139,7 +139,7 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             client_path = url_path_join(context_path, proxied_path)
         else:
             client_path = proxied_path
-        
+
         client_path = quote(client_path)
 
         client_uri = '{protocol}://{host}:{port}{path}'.format(
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -139,6 +139,8 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             client_path = url_path_join(context_path, proxied_path)
         else:
             client_path = proxied_path
+        
+        client_path = quote(client_path)
 
         client_uri = '{protocol}://{host}:{port}{path}'.format(
             protocol=protocol,
@@ -155,7 +157,7 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
 
         headers = self.proxy_request_headers()
 
-        client_uri = self.get_client_uri('http', host, port, quote(proxied_path))
+        client_uri = self.get_client_uri('http', host, port, proxied_path)
         # Some applications check X-Forwarded-Context and X-ProxyContextPath
         # headers to see if and where they are being proxied from.
         if not self.absolute_url:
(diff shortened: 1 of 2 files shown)


manics:
A couple of questions which are unclear to me:
- Are there ever any cases where `proxied_path` shouldn't be quoted?
- Should all calls to `get_client_uri` quote `proxied_path`, or only some?

oeway:
I did some quick search and here is what I think: the proxied path from tornado server is a decoded path, in all cases; and what required by the http client is a valid url, meaning it should always be encoded (quoted). Here is a thread about the httpclient https://github.com/tornadoweb/tornado/issues/559 . They also mentioned that the quote should not be called twice, which I think won’t happen here because the url/path is decoded already. 

Others may also comment , but I would conclude this can be applied to all cases.

manics:
Thanks for checking! `get_client_uri` is also called in `proxy_open`:
https://github.com/jupyterhub/jupyter-server-proxy/blob/80263c5748f1e0da19fe6b54e6b11883d141788f/jupyter_server_proxy/handlers.py#L254

It would also be great if you could add a test https://github.com/jupyterhub/jupyter-server-proxy/blob/master/tests/test_proxies.py. The tests [use an echo-style server](https://github.com/jupyterhub/jupyter-server-proxy/blob/master/tests/resources/httpinfo.py) so it should be straightforward

Might be worth seeing if anyone else has comments first.

oeway:
Hi @manics , thanks for pointing out, I think the same should also apply to websocket (`proxy_open`) so I moved the quote function to `get_client_uri`.

I added a test, let me know if you have better idea how to test it. As you will see from the commit history, I did some experiment to show the fix is actually works with the test I added.

BTW, although the tests passed, there is [an error trace back in the history](https://travis-ci.org/github/jupyterhub/jupyter-server-proxy/builds/662550007#L373-L391), is that normal?

Issue #147, opened 2019-09-21
Windows support


jacobtomlinson:
When trying to run [jupyterlab-nvdashboard](https://github.com/jacobtomlinson/jupyterlab-nvdashboard) on windows I'm getting the following error on startup.

![image](https://user-images.githubusercontent.com/1610850/65372592-409a2a00-dc6a-11e9-8551-662341565cc5.png)
_(Sorry about the screenshot, copying from powershell seems to lose all linebreaks)_

I imagine this is related to the way the server proxy is using asyncio/tornado to create subprocesses.

**Conda env**
```
name: pipeline
channels:
  - conda-forge
  - defaults
dependencies:
  - attrs=19.1.0=py_0
  - backcall=0.1.0=py_0
  - bleach=3.1.0=py_0
  - bokeh=1.3.4=py37_0
  - ca-certificates=2019.5.15=1
  - certifi=2019.6.16=py37_1
  - click=7.0=py_0
  - cloudpickle=1.2.2=py_0
  - colorama=0.4.1=py_0
  - cudatoolkit=10.0.130=0
  - cudnn=7.6.0=cuda10.0_0
  - cupy=6.0.0=py37h230ac6f_0
  - cytoolz=0.10.0=py37hfa6e2cd_0
  - dask=2.4.0=py_0
  - dask-core=2.4.0=py_0
  - decorator=4.4.0=py_0
  - defusedxml=0.5.0=py_1
  - distributed=2.4.0=py_0
  - entrypoints=0.3=py37_1000
  - fastrlock=0.4=py37h6538335_1000
  - freetype=2.10.0=h563cfd7_1
  - fsspec=0.5.1=py_0
  - heapdict=1.0.0=py37_1000
  - intel-openmp=2019.5=281
  - ipykernel=5.1.2=py37h5ca1d4c_0
  - ipython=7.8.0=py37h5ca1d4c_0
  - ipython_genutils=0.2.0=py_1
  - jedi=0.15.1=py37_0
  - jinja2=2.10.1=py_0
  - jpeg=9c=hfa6e2cd_1001
  - json5=0.8.5=py_0
  - jsonschema=3.0.2=py37_0
  - jupyter_client=5.3.1=py_0
  - jupyter_core=4.4.0=py_0
  - jupyterlab=1.1.4=py_0
  - jupyterlab_server=1.0.6=py_0
  - libblas=3.8.0=12_mkl
  - libcblas=3.8.0=12_mkl
[...]

PR #181, opened 2020-03-24
Fix SupervisedProcess on Windows

oeway:
This PR fixes https://github.com/jupyterhub/jupyter-server-proxy/issues/147 by running another even loop (`ProactorEventLoop`) which support subprocess on Windows.

(Tested on Windows 10)


jupyter_server_proxy/handlers.py
```diff
@@ -444,7 +446,8 @@ def get_timeout(self):
         return 5
 
     async def _http_ready_func(self, p):
-        url = 'http://localhost:{}'.format(self.port)
+        # Use 127.0.0.1 instead of localhost, since it can fail on windows
+        url = 'http://127.0.0.1:{}'.format(self.port)
```
manics: Could a few others review this change? I've run into problems with `localhost` vs `127.0.0.1` before on Linux, so we should be careful not to inadvertently break someone else. For example how does this behave with IPv6? Should we actually make this configurable or is that adding unnecessary complexity?

jupyter_server_proxy/handlers.py
```diff
@@ -457,6 +460,32 @@ def get_timeout(self):
...
     async def ensure_process(self):
+        """
+        Start the process
+        """
+        # The default asyncio event loop implementation on Windows does not 
+        # support subprocesses. Subprocesses are available for Windows if a 
+        # ProactorEventLoop is used. 
+        # Ref: https://docs.python.org/3/library/asyncio-subprocess.html
+        # We use a workaround for windows by starting a new ProactorEventLoop
+        # in a thread. This is necessary because we don't want it to interfer with
```
manics:
```suggestion
        # in a thread. This is necessary because we don't want it to interfere with
```

oeway:
> Thanks for investigating this. Do you have any experience of using Travis on Windows? https://docs.travis-ci.com/user/reference/windows/
> 
> I'm worried this could be inadvertently broken in future since most people develop and test on Linux or OS X.

No, but I just did a quick search and found travis does not support python on windows or OSX:
> Python builds are not available on the macOS and Windows environments.

https://docs.travis-ci.com/user/languages/python/


Would it be ok to switch to Github Actions? [All 3 major OS are supported](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#specifying-a-python-version), I can potentially give a quick try.

jupyter_server_proxy/handlers.py
```diff
@@ -457,6 +460,32 @@ def get_timeout(self):
...
+                new_loop.add_signal_handler = add_signal_handler
+                asyncio.set_event_loop(new_loop)
+                new_loop.run_until_complete(self._ensure_process())
+                new_loop.close()
+            loop = asyncio.get_event_loop()
+            await loop.run_in_executor(None, start_thread)
+        else:
+            await self._ensure_process()
+
+    async def _ensure_process(self):
```
yuvipanda: `_ensure_process` can now be called either from the main thread (where all the other couroutines are called from), or from an entirely different, ephemeral thread with its own event loop (which will only call _ensure_process). This might need some amount of cross-thread synchronization - for example, `self.state` could now be called from multiple threads. Does this sound correct?

I'm not actually sure how to go from here - my experience with python multi-threading is very limited, and mixing multi-threading with two event loops is even more limited.

jupyter_server_proxy/handlers.py
```diff
@@ -457,6 +460,32 @@ def get_timeout(self):
...
+        # ProactorEventLoop is used. 
+        # Ref: https://docs.python.org/3/library/asyncio-subprocess.html
+        # We use a workaround for windows by starting a new ProactorEventLoop
+        # in a thread. This is necessary because we don't want it to interfere with
+        # the event loop for tornado web server (which does not support ProactorEventLoop).
+        if sys.platform == "win32":
+            def start_thread():
+                new_loop = asyncio.ProactorEventLoop()
+                # patch add_signal_handler for simpervisor
+                def add_signal_handler(signum, callback, *args):
```
yuvipanda: Simpervisor needs these to watch for child processes dying so it can restart them, which is the core purpose of simpervisor. I'm guessing that doesn't work on Windows? If so, we might need an entirely different implementation in simpervisor for windows to work...

oeway:
@manics Made a PR for Github Actions which supports windows: https://github.com/jupyterhub/jupyter-server-proxy/pull/184

jupyter_server_proxy/handlers.py
oeway: The `ProactorEventLoop` on windows has no implementation for `add_signal_handler`.

jupyter_server_proxy/handlers.py
oeway: In our particular case, the `self.state['proc_lock']` and `self.state['proc']` is only used in the ensure_process thread, we will be completely fine. It will be an issue if multiple thread starts to write it, and we will need to lock the state.

oeway:
> Thank you very much for opening this PR! Windows support is absolutely necessary for many many tools these days, and while I don't have access to a machine, am happy there is work being done here!
> 
> However, simpervisor is fairly Linux / OS X specific - particularly, it relies on [catching SIGCHLD signal](https://docs.oracle.com/cd/E19455-01/806-4750/signals-7/index.html) to notice when the spawned process has died, and restart it. I'm not sure what a similar mechanism in Windows, but the most likely way forward is to implement that in Simpervisor (or a similar library), and use that here. I recognize that is way more work, but that's probably the way forward - mocking out signal handlers mean simpervisor can't do its job.

Thank you! I am happy that you commented here.

Indeed, this PR is more about make the subprocess run on windows, but I didn't try to test or fix the supervision part.

Since there is no POSIX signal on Windows, as you said, the supervision won't work. And we may need completely different implementation for the supervision.

oeway:
I did some search and couldn't find a solution for making sure process on windows are not dead. Well perhaps we don't need simpervisor if we already start the subprocess in a different thread? We can simply wait for the process to stop and restart again:
```python
while True:
    proc = SupervisedProcess(...)
    await proc.start()
    is_ready = await proc.ready()
    if not is_ready:
        await proc.kill()
        raise web.HTTPError(500, 'could not start {} in time'.format(self.name))
    process.wait() # this will block the process until it stops
```
What do you think? 
@yuvipanda

oeway:
A gentle ping to @yuvipanda : Could you please let me know what you think about how to support windows?

oeway:
FYI: Since this PR is suspended, I had to publish the patched version for windows on pypi: https://pypi.org/project/jupyter-server-proxy-windows/ . It's a big urgly, but if you want a quick solution for windows users, do `pip install jupyter-server-proxy-windows`.

consideRatio:
Thank you for pionieering work towards supporting windows @oeway!!!

Support for windows has now been added by introducing it in Simpervisor (https://github.com/jupyterhub/simpervisor/pull/26), allowing this project to not do much besides start testing against Windows as I understand it (#392).

Issue #182, opened 2020-03-25
Increase the request timeout


oeway:
Hi, I am use the server proxy to load some url with heavy data loads, and constantly encounter request timeout error like the one shown below:
```bash
[E 20:04:41.853 NotebookApp] Uncaught exception GET /elfinder/connector/?cmd=open&target=1046580b29204cf46ab9ae1f3a108d23&_=1585163057790 (127.0.0.1)
    HTTPServerRequest(protocol='http', host='xxxxxxxxxxxxxxxx', method='GET', uri='/elfinder/connector/?cmd=open&target=1046580b29204cf46ab9ae1f3a108d23&_=1585163057790', ve
rsion='HTTP/1.1', remote_ip='127.0.0.1')
    Traceback (most recent call last):
      File "/home/user/miniconda3/lib/python3.7/site-packages/tornado/web.py", line 1703, in _execute
        result = await result
      File "/home/user/miniconda3/lib/python3.7/site-packages/jupyter_server_proxy/websocket.py", line 96, in get
        return await self.http_get(*args, **kwargs)
      File "/home/user/miniconda3/lib/python3.7/site-packages/jupyter_server_proxy/handlers.py", line 507, in http_get
        return await self.proxy(self.port, path)
      File "/home/user/miniconda3/lib/python3.7/site-packages/jupyter_server_proxy/handlers.py", line 503, in proxy
        return await super().proxy(self.port, path)
      File "/home/user/miniconda3/lib/python3.7/site-packages/jupyter_server_proxy/handlers.py", line 213, in proxy
        response = await client.fetch(req, raise_error=False)
    tornado.simple_httpclient.HTTPTimeoutError: Timeout during request
```
It seems the default timeout for requests in tornado client is [20s](https://www.tornadoweb.org/en/stable/httpclient.html#request-objects), which is much shorter than most of the OS (interestingly it's OS specific and not related to browser itself) except for Windows. [Here](https://www.stevesouders.com/blog/2014/11/14/request-timeout/) is a nice blog post with lots of data in a [spreadsheet](https://docs.google.com/spreadsheets/d/1qBwgixmekvlvQH8V0EK5BWM9j0U9QanGa-PR03lwmKU).

[...]

PR #183, opened 2020-03-25
Capture request timeout error and increase the timeout value

oeway:
This PR aims to fix https://github.com/jupyterhub/jupyter-server-proxy/issues/182, it:
1. Increase the `connect_timeout=250.0, request_timeout=300.0` (both defaults to 20s) (these two values are used as default for FireFox).
2. Capture the timeout error, return `599: Timeout during request` rather than `500: Internal Server Error`.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -217,6 +217,10 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
         try:
             response = await client.fetch(req, raise_error=False)
         except httpclient.HTTPError as err:
+            # We need to capture the timeout error even with raise_error=False,
+            # because it only affects the HTTPError raised when a non-200 response 
+            # code is used, instead of suppressing all errors.
+            # Ref: https://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.AsyncHTTPClient.fetch
             if err.code == 599:
                 self._record_activity()
                 self.set_status(599)
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -308,7 +308,7 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
     def proxy_request_options(self):
         '''A dictionary of options to be used when constructing
         a tornado.httpclient.HTTPRequest instance for the proxy request.'''
-        return dict(follow_redirects=False)
+        return dict(follow_redirects=False, request_timeout=75)
 
     def check_xsrf_cookie(self):
         '''
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -11,7 +11,7 @@ from urllib.parse import urlunparse, urlparse, quote
 import aiohttp
 from asyncio import Lock
 
-from tornado import gen, web, httpclient, httputil, process, websocket, ioloop, version_info
+from tornado import gen, web, httpclient, simple_httpclient, httputil, process, websocket, ioloop, version_info
 
 from notebook.utils import url_path_join
 from notebook.base.handlers import IPythonHandler, utcnow
@@ -213,7 +213,15 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
         client = httpclient.AsyncHTTPClient()
 
         req = self._build_proxy_request(host, port, proxied_path, body)
-        response = await client.fetch(req, raise_error=False)
+
+        try:
+            response = await client.fetch(req, raise_error=False)
+        except simple_httpclient.HTTPTimeoutError as err:
+            self._record_activity()
+            self.set_status(408)
+            self.write(str(err))
+            return
+
         # record activity at start and end of requests
         self._record_activity()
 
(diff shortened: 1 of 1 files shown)


ryanlovett:
Thanks for investigating and the PR! Your logic seems sound to me. Just one comment...

I'd prefer if we catch tornado.httpclient.HTTPClientError, check whether its .code is a 599, and respond in the way you've written, raising if otherwise. Tornado chooses the simple client by default, but we us the base httpclient interface all throughout and it is common to both tornado's simple and curl http clients. (not that we provide a way to use the curl client instead)

What do you think? Anyone else have a comment?

oeway:
@ryanlovett Although I am not familiar with the project enough, I think that make sense and I have changed according to your suggestion.

In the meantime, I did a bit more investigation on the connection timeout for browsers, Firefox (on Mac) for example has the following settings (can be changed via `about:config`):
`network.http.response.timeout=300`
`network.http.connection-timeout=90`
`network.http.connection-retry-timeout=250`

Thus, I changed the `connect_timeout` for the HttpClient to 250s, and also increased the `request_timeout` to 300s. I think it's safe and better to use a large value for both, since we don't want the proxy client to give up a lot earlier than the browser.

jupyter_server_proxy/handlers.py
```diff
@@ -213,7 +213,18 @@ def _check_host_whitelist(self, host):
         client = httpclient.AsyncHTTPClient()
 
         req = self._build_proxy_request(host, port, proxied_path, body)
-        response = await client.fetch(req, raise_error=False)
+
+        try:
+            response = await client.fetch(req, raise_error=False)
+        except httpclient.HTTPError as err:
```
betatim: The `client.fetch` call uses `raise_error=False`. I think this means no exceptions will be raised. Is this correct or is the argument confusingly named (no exceptions except for timeouts)?

If there are no exceptions we should probably deal with timeouts in the code below

oeway: We do see the timeout exception even with `raise_error=False`, see the stack trace here: https://github.com/jupyterhub/jupyter-server-proxy/issues/182.

In the [docs](https://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.AsyncHTTPClient.fetch), it said explicitly it will not suppress all errors:
```
The raise_error=False argument only affects the HTTPError raised when a non-200 response code is used, instead of suppressing all errors.
```

So why the timeout error is not `HTTPError` and being suppressed? I guess if there is a timeout, then there won't be any response, thus the error cannot be suppressed, I think.

betatim: Aha! Thanks for digging out the docs :)

Maybe we can add a comment to the code which contains your response/link to docs to remind future generations that this seemingly weird setup is on purpose.

oeway: Good idea, done!

ryanlovett: @betatim I think raise_error=False tells the client to return the response to the caller, regardless of error code, rather than raise an exception. I think we want to always get the response, but we could conceivably return other things to the user based on the http code if we wanted.

Also, in tornado <= 5.1 it would [set the return code for errors to 599](https://www.tornadoweb.org/en/branch5.1/httpclient.html#tornado.httpclient.AsyncHTTPClient.fetch). If people are still using 5.1 with this proxy and this PR, they may start to see a lot of timeout errors whereas before we would just return the response without checking the error code. I don't think we care too much but it's worth noting. If a lot of people are still on 5.1, I think we'd have to insert a check on the tornado version and use the old behavior for <= 5.1.

oeway: @ryanlovett Could you please clarify why `they may start to see a lot of timeout errors whereas before we would just return the response without checking the error code.`?  

To be clear, the only thing we did in this PR is to capture the error which previously treated as internal server error (we are not changing the setting of `raise_error=False`). In the link you mentioned, it said:
> The raise_error=False argument currently suppresses all errors, encapsulating them in HTTPResponse objects with a 599 response code. This will change in Tornado 6.0: raise_error=False will only affect the HTTPError raised when a non-200 response code is used.

If I understand correctly, tornado <=5.1 won't raise any error, so the code should function exactly as before. And for tornado >5.1 (I am testing with 6.0+), it will raise the error then we will start to return timeout errors instead of saying the internal server error.

ryanlovett: We won't see more exceptions due to raise_error=False, but when suppressing the exceptions, tornado <= 5.1 would set the return code to be 599. This PR examines that code... only when there is an exception. So... never mind. Sorry.

rsignell-usgs: This is exciting guys!  Can't wait to try this out!  Thanks @oeway!

ryanlovett: @betatim Any final thoughts before merging?

betatim: No thoughts. My confusion was from the "split" behaviour of the new meaning of the argument. That was addressed in https://github.com/jupyterhub/jupyter-server-proxy/pull/183#discussion_r398504580. Marking this conversation as resolved.

betatim:
Thanks for working on this and the education about the new behaviour of the "no exceptions please" flag :)

PR #187, opened 2020-03-31
Update readme with labextension info

consideRatio:
There was missing information on the JupyterLab extension which this PR fixes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,7 +88,7 @@ Python package
 
    pip install -e .
 
-   # explicit install needed with editable mode (``-e``) jupyter
+   # explicit install needed with editable mode (-e) jupyter
    serverextension enable --sys-prefix jupyter_server_proxy
 
 
--- a/README.rst
+++ b/README.rst
@@ -20,8 +20,12 @@ Jupyter Server Proxy
    :target: https://www.npmjs.com/package/@jupyterlab/server-proxy
 
 Jupyter Server Proxy lets you run arbitrary external processes (such
-as RStudio, Shiny Server, syncthing, PostgreSQL, etc) alongside your
-notebook, and provide authenticated web access to them.
+as RStudio, Shiny Server, Syncthing, PostgreSQL, Code Server, etc)
+alongside your notebook server and provide authenticated web access to
+them using a path like ``/rstudio`` next to others like ``/lab``.
+Alongside the python package that provides the main functionality, the
+JupyterLab extension (``@jupyterlab/server-proxy``) provides buttons
+in the JupyterLab launcher window to get to RStudio for example.
 
 **Note:** This project used to be called **nbserverproxy**. As
 nbserverproxy is an older version of jupyter-server-proxy, uninstall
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -20,8 +20,12 @@ Jupyter Server Proxy
    :target: https://www.npmjs.com/package/@jupyterlab/server-proxy
 
 Jupyter Server Proxy lets you run arbitrary external processes (such
-as RStudio, Shiny Server, syncthing, PostgreSQL, etc) alongside your
-notebook, and provide authenticated web access to them.
+as RStudio, Shiny Server, Syncthing, PostgreSQL, Code Server, etc)
+alongside your notebook server and provide authenticated web access to
+them using a path like `/rstudio` next to others like `/lab`.
+Alongisde the python package that provide the main functionality, the
```
manics, line 26:
```suggestion
Alongside the python package that provides the main functionality, the
```

README.rst
```diff
@@ -20,8 +20,12 @@ Jupyter Server Proxy
...
+JupyterLab extension (`@jupyterlab/server-proxy`) provide buttons in
```
manics, line 27:
```suggestion
JupyterLab extension (`@jupyterlab/server-proxy`) provides buttons in
```
This is an RST file so I don't think the single backticks do what you intended (format as code?). You could try double backticks but run a sphinx/rst build to verify.

README.rst
```diff
@@ -43,33 +47,75 @@ The primary use cases are:
...
+^^^^^
 
 .. code-block::
 
    conda install jupyter-server-proxy -c conda-forge
 
-src
----
+JupyterLab extension
+--------------------
```
manics, line 72: Since you're updating this, do you think it's worth mentioning you also need to pip-install the python package too, or is this already clear enough?

consideRatio:
```suggestion

Note that as the JupyterLab extension only is a graphical interface to launch registered applications in the python package, the extension requires the python package to be installed.

```

consideRatio:
Thanks @manics for the code suggestions!

manics:
:+1:
There was one more comment which I didn't make suggestions for because I didn't have time to test:
> This is an RST file so I don't think the single backticks do what you intended (format as code?). You could try double backticks but run a sphinx/rst build to verify.

consideRatio:
@manics I've failed to review the README.rst as it is not part of the docs folder and my RST skills are very low. I think I got it right by looking at gitlab preview: https://github.com/jupyterhub/jupyter-server-proxy/blob/9b5d51a9ccfc4ec8917f6b0f13c274e4860dddb6/README.rst

README.rst
```diff
@@ -43,33 +47,79 @@ The primary use cases are:
...
+============
+
+Python package
+--------------
+
+.. code-block::
+
+   pip install -e .
+
+   # explicit install needed with editable mode (``-e``) jupyter
```
manics, line 91:
```suggestion
   # explicit install needed with editable mode (-e) jupyter
```
This is already inside a code-block:
image

PR #189, opened 2020-04-24
Fix expected type of ServerProxy command callable

cmd-ntrf:
ServerProxy command if callable should return a list, while in the docstring it says it should return a dictionary.


yuvipanda:
Good catch, @cmd-ntrf! Thanks :)

PR #190, opened 2020-04-24
Rename server_proccesses to server_processes

cmd-ntrf:
The extra c got my OCD tingling.


yuvipanda:
Haha I never can get these right. Thanks for catching it.

PR #191, opened 2020-04-24
Set list item class as sever_process.name in tree.js

cmd-ntrf:
All server process entries shared the same class named after rstudio.


yuvipanda:
Hah! Another great catch, a holdover from the original version from @ryanlovett. Thanks for catching it!

PR #192, opened 2020-04-27
In JupyterLab, add proxy to command registry even if launcher entry is disabled

cmd-ntrf:
Currently, when a server proxy launcher is entry is not enabled in the configuration file, the server proxy command is not added to the command registry. Other extensions could be interested in handling the launcher entries while jupyterlab-server-proxy handles the registration of the commands.

This is the case for [jupyterlab-lmod extension](https://github.com/cmd-ntrf/jupyter-lmod), where we want to show the launcher button only when the corresponding software module is loaded. I have made a short video of the workflow https://youtu.be/EQOGcoK2mSE. In order to achieve what is demonstrated, I had to copy the code for the command creation and launcher entry adding from jupyter-server-proxy in jupyter-lmod.

### DIFF
--- a/jupyterlab-server-proxy/src/index.ts
+++ b/jupyterlab-server-proxy/src/index.ts
@@ -7,11 +7,6 @@ import '../style/index.css';
 
 function addLauncherEntries(serverData: any, launcher: ILauncher, app: JupyterFrontEnd) {
     for (let server_process of serverData.server_processes) {
-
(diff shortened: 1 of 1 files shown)


ryanlovett:
This seems useful to me, thanks!

ryanlovett:
@cmd-ntrf I'll merge this unless you're still working on it. Just let me know.

cmd-ntrf:
Good to merge.

PR #193, opened 2020-04-27
Add ability to open proxied servers in JupyterLab tabs

cmd-ntrf:
Aim to render the proxied server in a JupyterLab tab instead of a blank browser tab when configured accordingly. The default is to keep opening a new browser tab.

I made a quick video of what it would look like: https://youtu.be/4H4xqQWLCW0

The proxied app has to allow to be framed. For example, for RStudio, the option `--www-frame-origin=same` has to be provided when launching rserver.

In its current state, there is no icon next to the tab name, it could be a nice to have.

Let me know what you think!


ryanlovett:
This looks really great! Thanks for this addition. I think this behavior matches up with lab users' expectations.

I wonder if people would get confused as to why some apps would launch in Lab and others would launch in a browser tab. If it ever became an issue, then maybe there could be a ui indicator when `framed` was false, since Lab normally opens things in frames.

Regarding, "the proxied app has to allow to be framed," do you have a pointer to relevant web standard or docs that describe this?

cmd-ntrf:
> I wonder if people would get confused as to why some apps would launch in Lab and others would launch in a browser tab. If it ever became an issue, then maybe there could be a ui indicator when `framed` was false, since Lab normally opens things in frames.

I agree. Not sure what would be the best indicator 

> Regarding, "the proxied app has to allow to be framed," do you have a pointer to relevant web standard or docs that describe this?

RStudio uses `X-Frame-Options`:
> The X-Frame-Options HTTP response header can be used to indicate whether or not a browser should be allowed to render a page in a <frame>, <iframe>, <embed> or <object>. Sites can use this to avoid clickjacking attacks, by ensuring that their content is not embedded into other sites.
Reference : https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

betatim:
Maybe add an icon like https://fontawesome.com/icons/external-link-alt?style=solid to the launcher text when it isn't being framed?

cmd-ntrf:
I like the idea of the icon.

The external link icon is currently used by JupyterLab as the icon for the launcher...
image. I don't know if it would create confusion the use the same icon to express something different.

cmd-ntrf:
I think this PR is now ready for review. I have made a quick video to demonstrate the evolution since the initial commit : https://youtu.be/VFmIy9Wb83k

A few remarks:
- In my initial commit, I implemented my own iFrame widget. After searching in JupyterLab code, there was already an IFrame widget, so I reused this one instead.
- The icon idea suggest by @betatim was not easy to implement given JupyterLab does not provide a lot of control on the appearance of the launcher card. I opted for a compromise by adding UTF-8 arrow symbol to the title to mean the command will launch in a new tab. I am open to suggestion.
- Since the proxies can opened in tab, I implemented the restorer logic. In case of a refresh or a shutdown, the tabs are restored as they were next time the JupyterLab workspace is accessed.
- To implement the restorer logic, I had to refactor the fetch + callback logic in activate, otherwise, the restoration would not work. I am unsure why. I am open to suggestion, but given the comment that were left in place regarding the callback, I think the change will be welcomed.
- The restorer logic was also simpler to implement by defining a single JupyterLab command with args to launch the proxied servers instead of defining one command per proxy.

Let me know what you think!

ryanlovett:
This looks amazing, thanks!

jupyterlab-server-proxy/src/index.ts
```diff
@@ -1,57 +1,115 @@
...
+  return widget;
+}
 
-function addLauncherEntries(serverData: any, launcher: ILauncher, app: JupyterFrontEnd) {
-    for (let server_process of serverData.server_processes) {
-      let commandId = 'server-proxy:' + server_process.name;
+async function activate(app: JupyterFrontEnd, launcher: ILauncher, restorer: ILayoutRestorer) : Promise<void> {
+  const response = await fetch(PageConfig.getBaseUrl() + 'server-proxy/servers-info');
+  if (!response.ok) {
+    console.log('Fetching metadata about registered failed. Make sure jupyter-server-proxy is installed');
```
ryanlovett, line 25: Did you intend the message to say "registered servers failed" ? Also with a period at the end of "installed".

ryanlovett: ... actually that message predates this PR, sorry!

cmd-ntrf:
One more thing: while looking for IFrame example in JupyterLab codebase, I found out that the help extension is using IFrame Widget to display readthedocs page. When they register help command, the command has an argument that make the help doc either open in a JupyterLab or a new browser tab. 

This is implemented here : 
this:
https://github.com/jupyterlab/jupyterlab/blob/f0153e0258b32674c9aec106383ddf7b618cebab/packages/help-extension/src/index.tsx#L364

Currently, this PR implemented a `framed` boolean while the help-extension has a `newBrowserTab` boolean. This PR `framed` boolean default value is false, help-extension `newBrowserTab` default value is false.

That said, my question is : should the default behavior for jupyterlab-server-proxy be to open a JupyterLab tab or a browser tab? To be consistent with other extensions, I think it should open a JupyterLab tab, and the `framed` variable be renamed to something like `newBrowserTab`. 

This is the complete opposite of what I originally implemented, so totally I am totally open to discuss this.

mboisson:
What drives the decision to open one app in a JupyterLab tab or in a new browser tab ? Assuming more and more options are provided as a button, I think it makes sense to launch everything in a JL tab. Could it be the default for everything, and somehow have an option to "detach" a JL tab into its own browser tab ?

cmd-ntrf:
> What drives the decision to open one app in a JupyterLab tab or in a new browser tab ? 

Some applications might have trouble or defend opening in an iframe for _security_ reason. This is the case for RStudio that requires a flag to allow being access through an iframe.

> Assuming more and more options are provided as a button, I think it makes sense to launch everything in a JL tab. Could it be the default for everything, and somehow have an option to "detach" a JL tab into its own browser tab ?

If the default is to open in a JupyterLab tab, the savvy user could always open a browser tab and enter the URL to the proxy. I am not aware of any mechanism to control the right-click menu for JupyterLab tab, so a detach option does not sound easy to implement.

ryanlovett:
> To be consistent with other extensions, I think it should open a JupyterLab tab, and the framed variable be renamed to something like newBrowserTab.

I agree.

If newBrowserTab defaults to false, we should make it clear to authors of dependent projects that the default value of this new parameter could potentially break the app they're proxying. Linking to the X-Frame-Options doc you provided would be helpful for them so they could check the source of their app if in doubt.

I don't have a sense as to what percentage of apps this would impact.

cmd-ntrf:
@ryanlovett : I have renamed `framed` to `new_browser_tab`, updated the docstring accordingly, and added a section to the server-process documentation on `new_browser_tab` where we refer to Mozilla X-Frame-Options docs and provide an example with RStudio.

Let me know if I missed something or left something unclear.

ryanlovett:
Tested and works perfectly, thanks!

I initially requested @yuvipanda's review since he wrote the typescript, but he reported he's very busy at the moment. Lgtm!

Issue #197, opened 2020-05-11
Sync versions of package and lab plugin


ryanlovett:
The [docs](https://jupyter-server-proxy.readthedocs.io/en/latest/contributing/release.html#release-checklist) say the lab plugin version should match the python package version:

> We keep these two versions in sync to avoid figuring out a matrix of ‘which version of the python package is compatible with which version of the JupyterLab plugin?’

Circa [support for lab 2.0](https://github.com/jupyterhub/jupyter-server-proxy/pull/174), the lab plugin got a bump to 2.x but the package remained behind at 1.x.

The next release of the package should rectify this, even if there is no leap in functionality.

Related: https://github.com/jupyterhub/jupyter-server-proxy/issues/175

consideRatio:
I think I favor having the npm plugin have its major version follow JupyterLab but having minor/patch versions for it released however we please. JupyterLab extensions It typically needs a major bump when JupyterLab bumps its major version, and having it match JupyterLab will probably be more helpful for all endusers.

Parts contributing to my bias to lock the npm package major version to JupyterLabs major version:
- the npm extension is released under the jupyterlab org, and extensions there typically bump its major along with JupyterLab, if only to signal they are with compatible with the new major version of jupyterlab.
- incrementing the major version in a npm package without actually introducing a breaking change because we made a major version bump of jupyter-server-proxy, adds some friction that isn't needed if there wasn't also a breaking change for the npm package
- symmetrically, if the npm package needs a major release, i don't think it makes sense for the python package to make a major release either - that would cause confusion and force users to read changelogs etc only to see there was no breaking change introduced etc.

I suggest we update the docs to indicate this instead - what do you think?

manics:
It's a balance between maintainability and user friendliness. In practice do you have any feeling for how often a breaking release of one component would be required without a release of the other? In the case of a new JupyterLab release if the NPM package follows semver it's only a major release if it's not backwards compatible. If the server package had a major release I'd have thought that implies a breaking API change or a major new feature, which would therefore require the front-end UI component to also change.

If we keep the separate versions do you have any thoughts on how to automate CI deployments? Options I can think of are:
- Prefix tags with the component (e.g. `pypy-1.2.3`, `npm-4.5.6`)
- Move the NPM package into a separate repo

consideRatio:
[...]

PR #198, opened 2020-05-14
Set `new_browser_tab` to False by default

yuvipanda:
IMO, this is a big breaking change that should be a
new major version. We should also try fix all the subprojects
that need fixing before we do this.

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -118,7 +118,7 @@ def make_server_process(name, server_process_config):
             icon_path=le.get('icon_path'),
             title=le.get('title', name)
         ),
-        new_browser_tab=server_process_config.get('new_browser_tab', False)
+        new_browser_tab=server_process_config.get('new_browser_tab', True)
     )
 
 class ServerProxy(Configurable):
@@ -176,7 +176,7 @@ class ServerProxy(Configurable):
               Title to be used for the launcher entry. Defaults to the name of the server if missing.
 
           new_browser_tab
-            Set to True to make the proxied server interface opened as a new browser tab. Set to False (default)
+            Set to True (default) to make the proxied server interface opened as a new browser tab. Set to False
             to have it open a new JupyterLab tab. This has no effect in classic notebook.
         """,
         config=True
@@ -192,7 +192,7 @@ class ServerProxy(Configurable):
         be proxied and False if it should not.  Such a function could verify
         that the host matches a particular regular expression pattern or falls
         into a specific subnet.  It should probably not be a slow check against
-        some external service.  Here is an example that could be placed in a 
+        some external service.  Here is an example that could be placed in a
         site-wide Jupyter notebook config:
 
             def host_whitelist(handler, host):


yuvipanda:
/cc @ryanlovett @cmd-ntrf @consideRatio 

Apologies for not reviewing the PR adding this earlier...

ryanlovett:
@yuvipanda Do you have a sense as to how many subprojects this breaks? The rstudio proxy has been patched, and shiny and vs-code don't need it. I'm sure there are several I'm not aware of.

In any case, a new major version would [help us sync](https://github.com/jupyterhub/jupyter-server-proxy/issues/197) the package and lab plugin versions. :)

yuvipanda:
Right, I think what we could do is:

1. Merge this, cut a new version that's a minor bump
2. Revert this, cut a new version that's a major bump

This way, we don't accidentally break people, since this *is* a major change. What do you think?

yuvipanda:
I am mostly worried about people with their own deployments, running their own applications with config.py files scattered around. This will break them, especially if they are using [framebusting](https://en.wikipedia.org/wiki/Framekiller).

yuvipanda:
I'd also personally think that 99% of apps here do *not* want to be iframed - it's an external facing app that someone wants to run under JupyterHub. Having very different behavior when launched from JupyterLab vs Classic Notebook is also very confusing IMO. So I'd still actually prefer for this to be opt-in rather than opt-out.

ryanlovett:
@cmd-ntrf How do you feel about @yuvipanda's concerns?

cmd-ntrf:
I think both options are valid, and I played with both when implementing PR #193, so I have nothing against reverting `new_browser_tab` to True as default. It makes sense given previous behavior of jupyter-server-proxy and wait for a major version release to change the default if this is what it is decided.

That said, on the long-term, I think the default behavior should be to open a JupyterLab tab.
- JupyterLab users except launcher items to start a new tab and as time goes by, there will be more users that are only familiar with JupyterLab, limiting the potential confusion. I do not think the extension behavior in the classic Notebook should be taken into account when making user-experience decision for the JupyterLab extension.
- There are other extensions proxying web app in JupyterLab, like [jupyterlab_tensorboard](https://github.com/chaoleili/jupyterlab_tensorboard), that open new JupyterLab tab. I think the demand for that sort of integration in JupyterLab is bigger than 1%. Being able to have a web app next to a notebook or a terminal has a lot of potential. RStudio or VS Code next to a notebook is a bit quirky, I agree, but I think there are more use cases that would benefit from being in a tab.
- The Wikipedia article on framebusting mentions the technique is idempotent when put in an HTML5 iframe with the sandbox attribute, which is what is used by JupyterLab iframe widget.

yuvipanda:
First, I'm extremely in awe that this functionality exists, and can be toggled on with a simple flag. Thank you very much for pushing this through, @cmd-ntrf!

Interestingly, I see jupyter-server-proxy's first goal as allowing arbitrary applications (like rstudio) to run under JupyterHub as first-class-citizens in full-screen ways. For example, at Berkeley, way more people use RStudio via this method than JupyterLab in any form. Classic notebook is what most people use to teach (at Berkeley and elsewhere, I think) and I don't think that's going away anytime soon. So I haven't really thought of the primary-JupyterLab use case so far.

Either way, I think change in the default behavior should definitely be a major version bump, not a minor. What do you think of that?

betatim:
I had seen the PR to add this functionality but was surprised when debugging my vscode binder (for unrelated reasons) to notice that vs code was now in a jupyterlab tab. I added code to https://github.com/betatim/vscode-binder/tree/master to make it open in a new browser tab as before. (As some feedback on which setting is "the right default value")

However I am always massively in favour of not changing existing behaviour for existing users, even if it is an improvement. It is super tedious for me as a maintainer to stick to that rule but Tim-the-user-of-software hates nothing more than stuff changing behaviour for "no good reason" (based on the fact that I was a happy user of the status quo :D )

manics:
_If_ it's the new default I think this is definitely a breaking change:
- it's a major UI change for users
- it breaks many existing configurations

cmd-ntrf:
2 nights of sleep later, I think this PR is a no-brainer.

I will make sure in the future to not break with default behavior when introducing new changes, and trust me, there is more to come ;).

betatim:
Thanks everyone for the fast and friendly resolution!

Issue #200, opened 2020-05-21
Unicode for newBrowserTab not displaying correctly


maresb:
### Bug description

The Unicode character 🡕 from [this line of code](https://github.com/jupyterhub/jupyter-server-proxy/blob/449f9d95db778296c96e82be272d929d0f9da5c8/jupyterlab-server-proxy/src/index.ts#L91) does not display in my browser. 

#### Expected behaviour

I expect to see the "North East Sans-Serif Arrow" character.

#### Actual behaviour

![Screenshot from 2020-05-21 22-40-32](https://user-images.githubusercontent.com/15216687/82604145-1eb7c500-9bb4-11ea-815a-2e212cd0c1dd.png)

### Your personal set up

 - OS:
Ubuntu 18.04.4
 - Version:
JupyterLab 2.1.1 running in Firefox 76.0.1

manics:
- Do you see the same problem when running on mybinder instead of locally? https://mybinder.org/v2/gh/betatim/vscode-binder/master?urlpath=lab definitely works for me
- Do you know if this is a problem with jupyter-server-proxy, or with jupyterlab? For instance if you install and create a kernel with a custom name:
https://ipython.readthedocs.io/en/latest/install/kernel_install.html#kernels-for-different-environments
Do you see the same problem?

maresb:
Hello @manics, things look identical on mybinder.

I'm not sure what you're getting at with the custom kernels, and I have to go to bed now, but I'll try to have a look tomorrow.

manics:
The idea of installing a custom kernel is you can set a custom unicode name, which should then appear in the Launcher. This will help to narrow down whether it's a unicode problem specific to jupyter-server-proxy, or if it's a more general problem with Jupyterlab.

ryanlovett:
I think this has to do with the unicode character from the new ["open in new tab" decoration](https://github.com/jupyterhub/jupyter-server-proxy/blob/master/jupyterlab-server-proxy/src/index.ts#L91), and probably isn't related to the launcher name or lab itself. The character from the screenshot is between brackets. If so, maybe the question is, what fonts are necessary to display that character?

Apparently the utf-8 arrow character was [easier to implement](https://github.com/jupyterhub/jupyter-server-proxy/pull/193#issuecomment-625400508) than inserting an icon.

PR #201, opened 2020-05-22
Use more common Unicode arrow

maresb:
The old character 🡕 is a relatively obscure sans-serif variant of the more standard arrow ↗, the later of which should have much better coverage.

This should solve #200.

If we want to do even better, several alternatives are available in [this SO question](https://stackoverflow.com/questions/1899772/new-window-icon-for-web-accessibility).


maresb:
Oops, I think I grabbed the wrong character...

PR #203, opened 2020-06-09
Bump version number

yuvipanda:


yuvipanda:
does this get auto deployed?

consideRatio:
Eh hehe no ;)

Tag commit and publish manually i think is current procedure for repo

PR #209, opened 2020-08-02
Add rewrite_response

### LINKED ISSUE #202: Inject a `<script>` tag into each HTML page
### Proposed change
Add ability to modify HTML as it is being proxied. For mybinder.org it would be cool (useful?) if we could inject something like `<script src="mybinder.org/static/bundle.js">` just before the closing `<body>` tag of every HTML page that is proxied. This would allow us to load some JS that then injects a "binder menu". This menu would contain links to the functionality that we currently only offer for classic notebook and (partly) jupyter lab. These are buttons that let people share the binder link, open a video conference, etc. A reason to have a JS bundle that does this is that it would be independent of the particular UI (voila, RStudio, RShiny, Jupyter) and would always appear. Right now we have to maintain two ways of doing this (one for lab, one for classic) and for RStudio users get nothing.

### Alternative options
Create a new package that derives from this one and implements the JS injection.

BinderHub operators could use this to show a small custom UI element and reduce maintenance burden on them because they don't have to create one for every frontend that they want to support. In turn all users of a BinderHub would use it because it gives them access to UX elements that otherwise are hidden from them.

### Suggest a solution
I would modify the handler for `GET` requests to check if the content type is HTML and if yes parse the HTML and inject a script tag as the last child of the body tag. The src URL for this bundle would be configured for the whole proxy.

I am not sure if we need to parse the HTML or if we could do a search and replace for `</body>` -> `<script src="..."></body>`. Parsing seems like it would be more robust but also slower??

Maybe there is a completely different way of achieving the same goal? What do people think of this?

maresb:
Hey all, here's what I came up with to resolve #202.

I seem to have achieved my original goal, but there's quite a lot I'm dissatisfied about.

1. In order to be able to read the config in the appropriate spots, I ended up having to change the signatures of several functions. Will this cause breakage in other projects? Is there some better way which I didn't see?
2. I couldn't make my new `rewrite_response` function take just `response` as input without losing important information. Thus I modified my original specification, and `rewrite_response` is now a function of `host`, `port`, `path`, and `response`.
3. I had to implement a near-duplicate function `non_service_rewrite_response` to handle the `/proxy/` endpoint. In this case I'm skeptical of the `jupyter-server-proxy` design. Namely, I would find it more intuitive to first define a correspondence between "j-s-p server" objects and endpoints, and then implement the `/proxy/` endpoint as a "j-s-p server". Was it a conscious design decision not to do this, or is it just the way things evolved?
4. I would like to be able to intercept redirects, but if I'm not mistaken, redirects typically have no response body. Thus my `rewrite_response` won't be of any help here.
5. I couldn't figure out how to rebuild the docs.

I do think this PR is a step forward. Regarding the points above, would it make sense to address any quick fixes, merge, and then make issues for any outstanding points?

### DIFF
--- a/jupyter_server_proxy/__init__.py
+++ b/jupyter_server_proxy/__init__.py
@@ -23,13 +23,19 @@ def _load_jupyter_server_extension(nbapp):
     base_url = nbapp.web_app.settings['base_url']
     serverproxy_config = ServerProxyConfig(parent=nbapp)
 
-    server_processes = [make_server_process(k, v) for k, v in serverproxy_config.servers.items()]
+    server_processes = [
+        make_server_process(name, server_process_config)
+        for name, server_process_config in serverproxy_config.servers.items()
+    ]
     server_processes += get_entrypoint_server_processes()
     server_handlers = make_handlers(base_url, server_processes)
     nbapp.web_app.add_handlers('.*', server_handlers)
 
-    # Set up default handler
-    setup_handlers(nbapp.web_app, serverproxy_config.host_allowlist)
+    # Set up default non-server handler
+    setup_handlers(
+        nbapp.web_app,
+        serverproxy_config.host_allowlist,
+    )
 
     icons = {}
     for sp in server_processes:
@@ -38,7 +44,7 @@ def _load_jupyter_server_extension(nbapp):
 
     nbapp.web_app.add_handlers('.*', [
         (ujoin(base_url, 'server-proxy/servers-info'), ServersInfoHandler, {'server_processes': server_processes}),
-        (ujoin(base_url, 'server-proxy/icon/(.*)'), IconHandler, {'icons': icons})
+        (ujoin(base_url, 'server-proxy/icon/(.*)'), IconHandler, {'icons': icons}),
     ])
 
 
--- a/jupyter_server_proxy/__init__.py
+++ b/jupyter_server_proxy/__init__.py
@@ -1,5 +1,5 @@
 from .handlers import setup_handlers
-from .config import ServerProxy, make_handlers, get_entrypoint_server_processes, make_server_process
+from .config import ServerProxy as ServerProxyConfig, make_handlers, get_entrypoint_server_processes, make_server_process
 from jupyter_server.utils import url_path_join as ujoin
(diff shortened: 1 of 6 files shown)


ryanlovett:
Hi @maresb ,

First, I'm very sorry for the delay, but thank you very much for the PR including the extra cleanups. Regarding your questions:

1. Possibly? Most downstream projects just use traitlets rather than subclassing. I don't know of a project which heavily relies on the latter. If we were using semver we could bump the major version to signify an API change, however we use the major version to reflect Lab compatibility. We could bump the minor version and mention breakage in the README or some such.
3. Things evolved this way. Originally there were no named servers.
4. We do have [request_headers_override](https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#request-headers-override) but not response_headers_override.

How would you feel about rebasing on master? I'd be happy to merge.

maresb:
@ryanlovett, wow, it's been a while since I thought about this! :laughing: I'm pleasantly surprised that you're interested in merging. I just rebased onto master.

Before rebase: 6e5313a2a6aece8073efc9753adf0538c83c0977
After rebase: ca0c5fd6a4cd242de8860817b8136c8818f0e717

Probably before merging I should have one more look over my commits and make sure they still do what was intended, and that I didn't make any errors during the rebase.

maresb:
BTW, what do you think about black/pre-commit? It may be nice to standardize the formatting, but then that messes with all the outstanding PRs. But then maybe one could use something like [this](https://gist.github.com/ndevenish/a7a9eb93cc19abbfe80f2daab16d3367) to rebase existing PRs.

ryanlovett:
Thanks @maresb , I really appreciate the rebase! Just let me know when you think it is ready.

I wouldn't mind having some sort of automatic formatting. Not having to make formatting decisions appeals to me. That can probably be proposed in a new issue and then other project members can chime in.

maresb:
@ryanlovett I'm fairly convinced that it's ready now. (I was worried that if I made a mistake with the rebase then I might inadvertently revert something.)

If it looks good to you, then please merge.

ryanlovett:
Thank you again @maresb !

Issue #211, opened 2020-08-26
Change subproto log to debug


bryevdv:
I'd like to propose changing this line:

https://github.com/jupyterhub/jupyter-server-proxy/blob/851eaf22ac5a82c61b441cfecc01b2bcefddca8a/jupyter_server_proxy/handlers.py#L347

To be `self.log.debug`. Currently e.g. for embedded Bokeh apps it spams the token sent in the subproto header to the console. This is fine (the token is not any secret), but is a bit excessive. It does not seem necessary at the "info" level (at least not the details—could also keep `info` just for the "Client sent subprotocols" part and `debug` for the rest). Happy to submit the PR if it would be accepted.

ryanlovett:
Setting to debug would be appropriate as I think that line was inserted solely for development. A PR would be great.

bryevdv:
Great, is that log under test somewhere that will also need updating?

ryanlovett:
Hmm, not sure what you mean by that. Which line?

bryevdv:
@ryanlovett Sorry my question was not clear. Is there a test that maintains that this log:
https://github.com/jupyterhub/jupyter-server-proxy/blob/851eaf22ac5a82c61b441cfecc01b2bcefddca8a/jupyter_server_proxy/handlers.py#L347
that it happens at all and/or that it happens at specifically the `info` level? If there is, then the PR would need to update that test at the same time, too. Asking because I am not familiar with his codebase or its tests.

ryanlovett:
Ahh, no, nothing depends on that line. Thanks for checking though!

PR #212, opened 2020-08-27
Change subproto logging to debug

bryevdv:
ref: #211 

Should note: I have not tested this (not sure what that would involve) and am assuming this is a standard logger with a `debug` method.

cc @ryanlovett


ryanlovett:
Thanks @bryevdv ! Yes, we're using [tornado's logging](https://www.tornadoweb.org/en/stable/log.html).

Issue #216, opened 2020-09-13
adding X_FORWARDED_PREFIX for wsgi compatible apps


gbrault:
In file jupyter-server-proxy/jupyter_server_proxy/handlers.py [in line ](https://github.com/jupyterhub/jupyter-server-proxy/blob/ee403c608f9b92ed712576e50be4c4539ca939a7/jupyter_server_proxy/handlers.py#L173) adding the following code

```
headers['X_FORWARDED_PREFIX'] = context_path
```

This allows for example flask/werkzeug app to be able to run seamless in jupyter lab context

adding the following lines in the app does the job from flask perspective:

```
from  werkzeug.middleware.proxy_fix import ProxyFix
....
After creating the Flask app:
app.wsgi_app = ProxyFix(app.wsgi_app, x_for=1, x_host=1, x_prefix=1)
```

With this, jypyter server proxy makes flask/werkzeug apps working seamlesly! 

And making sure the front proxy is passing the appropriate headers, with nginx for example
x-for, x-host
```
real_ip_recursive on;

server {
  client_max_body_size 200M;
  server_name jod.seadev.xyz;
  
  add_header X-Forwarded-For $proxy_add_x_forwarded_for;
  add_header X-Remote-IP $remote_addr;
  add_header X-Forwarded-Host jod.seadev.xyz;

  location / {
    proxy_set_header        Host $host;
    proxy_set_header        X-Forwarded-Host jod.seadev.xyz;
    proxy_set_header        X-Real-IP $remote_addr;
    proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header        X-Forwarded-Proto $scheme;
    proxy_pass              http://127.0.0.1:8888;
    # WebSocket support (nginx 1.4)
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "upgrade";
    proxy_read_timeout  36000s;    
  }
}
```

I can send a pull request if you agree?

yuvipanda:
Thank you for digging into this! Would love a pull request, this would indeed be very useful!

ryanlovett:
Fixed by #217.

PR #217, opened 2020-09-14
pull request for "adding X_FORWARDED_PREFIX for wsgi compatible apps #216"

gbrault:
This enable to have flask/werkzeug apps seamless integration: see issue #216

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -173,7 +173,6 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             headers['X-Forwarded-Context'] = context_path
             headers['X-ProxyContextPath'] = context_path
             # to be compatible with flask/werkzeug wsgi applications
-            # headers['X_FORWARDED_PREFIX'] = context_path
             headers['X-Forwarded-Prefix'] = context_path
 
         req = httpclient.HTTPRequest(
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -172,6 +172,7 @@ class ProxyHandler(WebSocketHandlerMixin, IPythonHandler):
             context_path = self._get_context_path(port)
             headers['X-Forwarded-Context'] = context_path
             headers['X-ProxyContextPath'] = context_path
+            headers['X_FORWARDED_PREFIX'] = context_path
 
         req = httpclient.HTTPRequest(
             client_uri, method=self.request.method, body=body,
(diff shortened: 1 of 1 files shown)


jupyter_server_proxy/handlers.py
```diff
@@ -172,6 +172,8 @@ def _build_proxy_request(self, host, port, proxied_path, body):
             context_path = self._get_context_path(port)
             headers['X-Forwarded-Context'] = context_path
             headers['X-ProxyContextPath'] = context_path
+            # to be compatible with flask/werkzeug wsgi applications
+            headers['X_FORWARDED_PREFIX'] = context_path
```
yuvipanda, line 176: Should this be 'X-Forwarded-Prefix'? That is the more common HTTP Header naming convention. https://werkzeug.palletsprojects.com/en/1.0.x/middleware/proxy_fix/ seems to say that, but not sure if I'm looking in the right part of the documentation.

yuvipanda (changes requested):
Thank you for opening this PR!

yuvipanda (changes requested):
Thank you for opening this PR!

jupyter_server_proxy/handlers.py
gbrault: you are right from the reference, but W3C says [Headers are case-insensitive](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html). do you want me to fix it?
Please let me now, I'll do another PR.
NB: I have a working case under test..

yuvipanda: Ah, I was more worried about the '_' instead of '-', rather than the case.

gbrault: Ok, my mistake, I verify!

gbrault: Funy enough, it's what I have (and working well) on my test platform: I make the change and verify it works.

gbrault: Just tested the change and it works as well! 

Do I make a new pull request?

jupyter_server_proxy/handlers.py
```diff
@@ -172,6 +172,9 @@ def _build_proxy_request(self, host, port, proxied_path, body):
             context_path = self._get_context_path(port)
             headers['X-Forwarded-Context'] = context_path
             headers['X-ProxyContextPath'] = context_path
+            # to be compatible with flask/werkzeug wsgi applications
+            # headers['X_FORWARDED_PREFIX'] = context_path
```
yuvipanda, line 176:
```suggestion
```

yuvipanda (commented):
Awesome! Let's just remove the commented out line, and I can merge this.

yuvipanda:
Thanks a lot for your patch, @gbrault! I hope this is useful for you, and you contribute more patches in the future! <3

gbrault:
Thanks guys!

PR #224, opened 2020-11-10
Use the recommended async with syntax to acquire and release an asyncio Lock

jasongrout:
I'm not sure exactly what is going on here, but we're seeing an error with this statement on Python 3.9:

```
      File "/[snip]/lib/python3.9/site-packages/jupyter_server_proxy/handlers.py", line 489, in ensure_process
        with (await self.state['proc_lock']):
    TypeError: object Lock can't be used in 'await' expression
```

See https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock

I think this also works in Python 3.5: https://docs.python.org/3/whatsnew/3.5.html#pep-492-coroutines-with-async-and-await-syntax

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -491,7 +491,7 @@ class SuperviseAndProxyHandler(LocalProxyHandler):
         # FIXME: Make sure this times out properly?
         # Invariant here should be: when lock isn't being held, either 'proc' is in state &
         # running, or not.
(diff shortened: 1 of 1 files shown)


jasongrout:
> Also, please make sure you followed the pull request template, as this will help us review your contribution more quickly.

I didn't see a template. When I created a PR, it just automatically filled in my commit message.

yuvipanda:
<3 thanks for this, @jasongrout!

PR #226, opened 2020-11-18
support to URL filename and URL parameters

jhgoebbert:
Hello everyone,

[jupyter-xprahtml5-proxy](https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy) adds support for [Xpra](https://xpra.org) (remote desktop) to Jupyter through the `jupyter-server-proxy` extension.

The [Xpra](https://xpra.org) HTML5 client can be configured through URL parameters. Expecially the username, password and encryption key is here of high importance to ensure that any user on a multiuser compute node can only connect to its own Xpra server. e.g. `http://localhost:10000/index.html?username=test&password=123456`

One might think, that this can be done through "mappath", but the Xpra-HTML5 client extracts the URL paramters from the URL through JavaScript in the browser. Therefor it requires that these URL parameters must not only be known on the server/python-side, but also on the client/JavaScript-side to be added to the URL of the launcher icon.

This is not supported yet by `jupyter-server-proxy` and added by this merge request.
The optional variable added to launcher_entry is called "urlfile".

How this can be used is shown in the development branch of jupyter-xprahtml5-proxy.
It now supports password+encryption, which was not possible before:
https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy/blob/v0.3.0_devel/jupyter_xprahtml5_proxy/__init__.py

Let me know, what you think.
I think it is important and hope to see this in jupyter-server-proxy soon.

Best,
Jens Henrik

### DIFF
--- a/jupyter_server_proxy/api.py
+++ b/jupyter_server_proxy/api.py
@@ -19,7 +19,8 @@ class ServersInfoHandler(IPythonHandler):
                 'name': sp.name,
                 'launcher_entry': {
                     'enabled': sp.launcher_entry.enabled,
-                    'title': sp.launcher_entry.title
+                    'title': sp.launcher_entry.title,
+                    'urlfile': sp.launcher_entry.urlfile,
                 },
                 'new_browser_tab' : sp.new_browser_tab
             }
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -99,7 +99,7 @@ def make_handlers(base_url, server_processes):
         ))
     return handlers
 
-LauncherEntry = namedtuple('LauncherEntry', ['enabled', 'icon_path', 'title'])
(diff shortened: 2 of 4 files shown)


jhgoebbert:
With much success we are using [jupyter-xprahtml5-proxy](https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy/tree/v0.3.0_devel) (enabled by this pull request) since a few weeks in production.  
It would be fantastic if this pull request could make it into jupyter-server-proxy, soon.

olifre:
Same here, we are also using this, right now in a short pilot phase, but it works very well for us and we plan to take it into production soon. It would be very nice to see this PR reviewed / accepted.

jhgoebbert:
Hello everyone,
I would like to see this PK merged. Does anyone has time to review it and give me suggestions how to improve it so it can get merged?

jhgoebbert:
A similar solution was now merged with `path_info`: https://github.com/jupyterhub/jupyter-server-proxy/pull/279

PR #229, opened 2020-12-30
Update dependencies to include jupyterlab 3.x.x - Fixes #219

### LINKED ISSUE #219: Support JupyterLab 3
With JupyterLab 3 release candidates out, it would be great to have a compatible `jupyter-server-proxy`.

dipanjank:
Changed jupyterlab component dependencies to include jupyterlab 3.0.0.

Fixes jupyterhub/jupyter-server-proxy#219


jtpio:
Thanks!

Maybe it could be useful to add a small CI step to build and install the lab extension, similar to: https://github.com/jupyterlab/jupyterlab_apod/blob/2df8db256d67a28b3f728e2c3b5adef0a7a67fa9/.github/workflows/build.yml#L1

dipanjank:
Sure, let me look into it.

jtpio:
A follow-up PR could also update the `setup.py` and the packaging of the extension, so it adopts the new distribution system.

Meaning that a single `pip install jupyter-server-proxy` would also install the JupyterLab 3.0 extension by default.

There is more information about the new distribution in the docs:

- https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#prebuilt-extensions
- https://jupyterlab.readthedocs.io/en/stable/extension/extension_tutorial.html

dipanjank:
Thanks for the resources. Yes, I know about that feature. Happy to work on
that as a follow-up but can we release the jupyterlab 3 compatible version
as-is? Or are you thinking of releasing both together?

jtpio:
Yes it would make sense to first release a new version of the extension to npm, with the changes from this PR, so it can be installed with JupyterLab 3.0 using the `jupyter labextension install`.

And then as a second step switch to the new distribution system, which will require releasing the Python package to PyPI.

dipanjank:
Yes, agreed.

dipanjank:
Okay, added the "build" job which installs the extension in a virtualenv with all supported python versions, installs jupyterlab, installs the extension.

jupyterlab-server-proxy/.gitignore
```diff
@@ -3,3 +3,5 @@ lib/
 node_modules/
 *.egg-info/
 .ipynb_checkpoints
+yarn.lock
+../.idea/
```
consideRatio, line 7:
```suggestion
../.idea/

```

consideRatio (approved):
This LGTM!

Along with merging this, I figure we should release a jupyterlab extension as a NPM package, but then transition to packaging it in Python. I opened #233 for that.

@dipanjank have you tried this to work locally as well btw? I'm not a very active maintainer on this project and have only now done a review of the changes rather than tried them.

@jtpio do you suggest we release v3.0.0 of the NPM package to indicate its JL 3 compatible, or stay with v2.x.y as it is not a breaking change involved?

dipanjank:
@consideRatio yes, I did the following

* Create a virtualenv and install jupyterlab.
* Install my local version with ``jupyter labextension install .``
* Install tensorflow (which comes bundled with tensorboard)
* Add a tensorboard proxy entry in ~/.jupyter/jupyter_notebook_config.json
* Start jupyterlab and verify that tensorboard is available at ``http://localhost:8888/tensorboard``.

I think this can be done as an automated test. Will give it a try at some point if you think it's valuable.

consideRatio:
@dipanjank nice, thank you!

consideRatio:
I released version 2.1.2 of the JupyterLab extension to NPM.

jtpio:
> do you suggest we release v3.0.0 of the NPM package to indicate its JL 3 compatible, or stay with v2.x.y as it is not a breaking change involved?

Usually semver is useful to communicate the type of changes (breaking or not) to the consumer of the package. For this extension if it's still compatible with both lab 2 and 3, then it should be fine to stick to `v2.x.y` I think.

ryanlovett:
See https://github.com/jupyterhub/jupyter-server-proxy/issues/197 for more discussion about npm package versioning.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/how-to-launch-browser-from-jupyterlab/7912/6

PR #236, opened 2021-01-24
Replace server-process list with linkable headings

manics:
The current arguments are documented as a list which means it's not possible to link to an anchor within the page.

For example, in https://github.com/jupyterhub/jupyter-rsession-proxy/issues/95#issuecomment-766348946 I wanted to link directly to `absolute_url`, but I could only link to the page.


yuvipanda:
Thanks, @manics!

PR #235, opened 2021-01-15
Add Jupyter Server extension data file

jtpio:
For the Jupyter Server `ServerApp`.

Also add a step on CI to check the extension is correctly installed.

Issue #233, opened 2021-01-10
Package the jupyterlab extension with the Python package


consideRatio:
As @jtpio describes in https://github.com/jupyterhub/jupyter-server-proxy/pull/229#issuecomment-755994606, we can package the jupyterlab extension to install automatically as part of installing the python package!

dipanjank:
I'd like to work on this, if not taken already!

consideRatio:
@dipanjank greatly appreciated! Please mention me for review, I'd like to learn about this procedure :)

jtpio:
Thanks all!

Linking to a couple of resources that might help get this started:

- Using the upgrade script: https://jupyterlab.readthedocs.io/en/stable/extension/extension_migration.html#upgrading-library-versions-using-the-upgrade-script
- Extension tutorial using the Python packaging for distributing the extension: https://jupyterlab.readthedocs.io/en/stable/extension/extension_tutorial.html
- Docs on the prebuilt extensions: https://jupyterlab.readthedocs.io/en/stable/extension/extension_dev.html#prebuilt-extensions

janjagusch:
Hi! I'd also like to give this a try. :)

I would suggest to take inspiration from [jupytext](https://github.com/mwouts/jupytext), where the JL extension has already been packages.

From what I could see I believe you need to
* add a `build:prod` script to the `package.json`
* change the `setup.py` to use `jupyter_packaging`

jtpio:
Nice, thanks @janjagusch!

Issue #109, opened 2019-03-13
NoneType can't be used in 'await' expression - May be harmless


jakirkham:
Ran into the following traceback when using dask-labextension with jupyter-server-proxy. This occurred with Tornado 6, but went away with Tornado 5. Am using JupyterLab 0.35.4, dask-labextension 0.3.1, Bokeh 1.0.4, and jupyter-server-proxy 1.0.0.

```python
LabApp - ERROR - Uncaught exception GET /proxy/9100/status (10.110.73.60)
HTTPServerRequest(protocol='http', host='10.31.241.45:9999', method='GET', uri='/proxy/9100/status', version='HTTP/1.1', remote_ip='10.110.73.60')
Traceback (most recent call last):
  File "/home/nfs/jkirkham/miniconda/envs/rapidsdev37/lib/python3.7/site-packages/tornado/web.py", line 1699, in _execute
    result = await result
  File "/home/nfs/jkirkham/miniconda/envs/rapidsdev37/lib/python3.7/site-packages/jupyter_server_proxy/websocket.py", line 87, in get
    return await self.http_get(*args, **kwargs)
  File "/home/nfs/jkirkham/miniconda/envs/rapidsdev37/lib/python3.7/site-packages/jupyter_server_proxy/handlers.py", line 247, in http_get
    return await self.proxy(port, proxy_path)
TypeError: object NoneType can't be used in 'await' expression
LabApp - ERROR - Uncaught exception GET /proxy/9100/status (10.110.73.60)
HTTPServerRequest(protocol='http', host='10.31.241.45:9999', method='GET', uri='/proxy/9100/status', version='HTTP/1.1', remote_ip='10.110.73.60')
Traceback (most recent call last):
  File "/home/nfs/jkirkham/miniconda/envs/rapidsdev37/lib/python3.7/site-packages/tornado/web.py", line 1699, in _execute
    result = await result
  File "/home/nfs/jkirkham/miniconda/envs/rapidsdev37/lib/python3.7/site-packages/jupyter_server_proxy/websocket.py", line 87, in get
    return await self.http_get(*args, **kwargs)
  File "/home/nfs/jkirkham/miniconda/envs/rapidsdev37/lib/python3.7/site-packages/jupyter_server_proxy/handlers.py", line 247, in http_get
    return await self.proxy(port, proxy_path)
TypeError: object NoneType can't be used in 'await' expression
```

ian-r-rose:
[...]

PR #238, opened 2021-02-02
Bump to 1.6.0 (setup.py) and add CHANGELOG.md

consideRatio:
After upgrading to using ServerApp I stopped being able to use /rstudio.

> When I was using NotebookApp, visiting `/user/erik/rstudio/` would work fine. Now when I'm using ServerApp and JupyterHub 1.3.0, it seems that visiting `/user/erik/rstudio/` redirects to `/auth-sign-in?appUri=%2F` which bounce to `/hub/auth-sign-in?appUri=%2F` which jupyterhub doesn't recognize and responds with 404.

This question in gitter jupyter/jupyter_server led to @blink1073 helping out!

> hi @consideRatio! It looks like jupyter-server-proxy needs a new release with jupyterhub/jupyter-server-proxy#235.

So, here is a PR to release 1.6.0 of the Python package. The JupyterLab extension is already updated since last changed, version 2.1.2.

## Review
- [ ] Added a CHANGELOG.md file
  I generated it with [github-activity](https://github.com/executablebooks/github-activity)
- [ ] Bumped setup.py version to 1.6.0 from 1.5.3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,23 @@
+## [1.6]
+
+### [1.6.0]
+
+This release adds support for JupyterLab 3.
+
+At this point, the Jupyterlab extension of version 2.1.2, needs to be installed
+alongside the Python package for JupyterLab launcher buttons to show up as the
+extension isn't yet bundled with the python package.
+
+#### Enhancements made
+
+* Add Jupyter Server extension data file (JupyterLab 3 support) [#235](https://github.com/jupyterhub/jupyter-server-proxy/pull/235) ([@jtpio](https://github.com/jtpio))
+* Update dependencies to include jupyterlab 3.x.x (JupyterLab 3 support) [#229](https://github.com/jupyterhub/jupyter-server-proxy/pull/229) ([@dipanjank](https://github.com/dipanjank))
+
+#### Documentation improvements
+
+* Replace server-process list with linkable headings [#236](https://github.com/jupyterhub/jupyter-server-proxy/pull/236) ([@manics](https://github.com/manics))
+* Rename the mamba-navigator example to gator in the documentation [#234](https://github.com/jupyterhub/jupyter-server-proxy/pull/234) ([@jtpio](https://github.com/jtpio))
+
+#### Contributors to this release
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

yuvipanda:
Thanks for opening this, @consideRatio!

I see that the *only* change is the setup.py bump. Is that all that's required?

Would this still continue working without jupyter_server when done?

consideRatio:
Npm package still needed, and no change needed to it beyond what was released ~2 weeks ago in 2.1.2. This is also said in the top of the changelog currently!

Packaging the jlab extension with the python package remains a feature to be implemented and is tracked by #233.

consideRatio:
Oh i see what you are asking now sorry, yes i think so. What's recently changed was a configuration of jupyter_server by default i think.

NotebookApp should still work i've assumed though, rather than manually tested with the current state of things.

consideRatio:
I saw an error logged as part of running the tests but didn't fail the test which worries me.

Screenshot_20210203-084558

maresb:
@consideRatio I have always constantly gotten these errors. It would be really great to clean things up and check everywhere that `None` is not awaited. (There are many places in the code where this is necessary.)

consideRatio:
Ah, I found https://github.com/jupyterhub/jupyter-server-proxy/issues/109, so this is a long standing seemingly harmless error. And, the tests runs successfully without jupyter-server installed. I opened https://github.com/jupyterhub/jupyter-server-proxy/issues/239 to reflect a lot of confusion about things I fail to disentangle at this point but to summarize: NotebookApp works.

consideRatio:
@yuvipanda the answer is YES, it would work without jupyter_server!

yuvipanda:
aaah, so this just makes a release for #235! Thanks for helping me understand, @consideRatio.

consideRatio:
@yuvipanda nice! 

Okay to cut 1.6.0?

yuvipanda:
@consideRatio feel free to :)

consideRatio:
Done!

PR #240, opened 2021-02-03
tests: remove bad test, add new clarifying current behavior

### LINKED ISSUE #239: One of our tests is malfunctioning
The test `test_server_proxy_minimal_proxy_path_encoding_complement` is giving a false positive. It is a consequence of a lot of confusion related to `#` and how it should be treated.

I've tried making something sensible, but I fail. If I can get a clear description of how `#` should be treated across the jupyter-server-proxy I could perhaps manage to fix it without just removing it which is a very sensible thing as well to do.

## Current understanding
`#` are anchors to navigate to within the HTML response that browsers typically don't send to the webserver, and everything following `#` may due to that not get proxied onwards if its passed to jupytser-server-proxy I think. Hmm... If `#` is passed, it can lead to something following it to be ignored, such as `?token=asdf`. But when is actually `?token=asdf` relevant? Hmm...

There is a) a browser, b) a jupyter server, c) jupyter-server-proxy, and d) the proxied-server. Where is the token query parameter we pass in our test suite consumed?

Yikes I just have too much confusion about all this in general to figure things out I think... Perhaps removing the test as it was created with confusion to begin with makes the most sense.

consideRatio:
We had a false positive test because it was `return`-ing before its assertions which was the actual test. With this updated test I make the current behavior clear, but I also document in the docstring a FIXME that this is only the current behavior verified with a test, not that it is the desired behavior which isn't known to me.

Closes #239

## Off topic
This is no fix for #109, but it highlights that you can get that kind of error when ?token=wrongtoken is passed to the Jupyter Notebook server, which probably is an issue with notebook server. Right?

### DIFF
--- a/tests/test_proxies.py
+++ b/tests/test_proxies.py
@@ -30,17 +30,6 @@ def test_server_proxy_minimal_proxy_path_encoding():
     s = r.read().decode('ascii')
     assert 'GET /{}&token='.format(special_path) in s
 
-def test_server_proxy_minimal_proxy_path_encoding_complement():
-    """Test that we don't encode ?# as a complement to the other test."""
-    test_url = '/python-http/?token={}#test'.format(TOKEN)
-    h = HTTPConnection('localhost', PORT, 10)
-    r = request_get(PORT, test_url, TOKEN)
-    h.request('GET', test_url)
-    return h.getresponse()
-    assert r.code == 200
-    s = r.read().decode('ascii')
-    assert 'GET /{}?token='.format(test_url) in s
-
 
 def test_server_proxy_non_absolute():
     r = request_get(PORT, '/python-http/abc', TOKEN)
--- a/tests/test_proxies.py
+++ b/tests/test_proxies.py
@@ -31,6 +31,59 @@ def test_server_proxy_minimal_proxy_path_encoding():
     assert 'GET /{}&token='.format(special_path) in s
 
 
+def test_server_proxy_hash_sign_encoding():
+    """
+    FIXME: This is a test to establish the current behavior, but if it should be
+           like this is a separate question not yet addressed.
+
+           Related: https://github.com/jupyterhub/jupyter-server-proxy/issues/109
+    """
+    h = HTTPConnection("localhost", PORT, 10)
+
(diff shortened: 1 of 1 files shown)


tests/test_proxies.py
```diff
@@ -23,25 +23,65 @@ def request_get(port, path, token, host='localhost'):
...
-    special_path = quote("Hello world 123 åäö 🎉你好世界±¥ :/[]@!$&'()*+,;=-._~", safe=":/?#[]@!$&'()*+,;=-._~")
-    # NOTE: we left out ?# as they would interact badly with our requests_get
-    # function's ability to pass the token query parameter.
+    special_path = quote("Hello world 123 åäö 🎉你好世界±¥ :/[]@!$&'()*+,;=-._~?key1=value1", safe=":/?#[]@!$&'()*+,;=-._~")
     test_url = '/python-http/' + special_path
     r = request_get(PORT, test_url, TOKEN)
     assert r.code == 200
     s = r.read().decode('ascii')
-    assert 'GET /{}?token='.format(special_path) in s
+    assert 'GET /{}&token='.format(special_path) in s
```
manics, line 31: This test was already a bit confusing due to the long list of unicode characters. Could you explain why the assertion needs to change? The related issue mentions `test_server_proxy_minimal_proxy_path_encoding_complement`.

consideRatio: the request_get() helper function returns the response to a request that is appended with the token as a query string, but the token is appended with `?token=...` or `&token=...` depending on if it found a `?` within the URL it is making a request to as it needs to do in order to not define the query params to start a second time, as indicated by the first `?` found in the URL.

manics: What's the reason for changing the URL in this test, as opposed to adding a new test? Are we losing a test case by changing it?

consideRatio: No, just making it more explicit that ? won't be escaped on arrival to the receiving end.

consideRatio:
@manics ok to merge this also? This is a cleanup of an erroneous test to a valid test describing the current behavior, and making a note that it may or may not be the desired behavior.

PR #241, opened 2021-02-03
docs: mv local dev setup to CONTRIBUTING.md and document tests

consideRatio:
I struggled a while with `pip install -e .` not functioning because I copied CI instructions from test.yaml because I didn't see any setup instructions in the CONTRIBUTING.md that we had. I found later that the README.md had instructions though.

So, I made a note in the README.md and test.yaml to look in CONTRIBUTING.md where I described also described how to run tests.


yuvipanda:
Thanks, @consideRatio

PR #245, opened 2021-02-10
Package jupyter lab extension

janjagusch:
Closes #233 

You can now install the package and the JL extension through `pip install -ve .`.

This change is breaking, as you have to remove Python 3.5 support.

PR #247, opened 2021-02-14
Add publish PyPI and NPM workflow

manics:
https://github.com/jupyterhub/jupyter-server-proxy/pull/245 incorporates the JupyterLab extension into the Python package and uses a single version number for both, which means when it's merged we can publish everything automatically.

This adds a new `Publish` workflow where the build and publish steps for PyPI and NPM are copied from
https://github.com/manics/jupyter-offlinenotebook/blob/340e0e7ac80f01d276c2e68d03144fe1078d5c86/.github/workflows/build.yml#L57-L72
https://github.com/manics/jupyter-offlinenotebook/blob/340e0e7ac80f01d276c2e68d03144fe1078d5c86/.github/workflows/build.yml#L118-L161

This requires two repository secrets:
- [x] `secrets.PYPI_PASSWORD`
- [x] `secrets.NPM_TOKEN`

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,81 @@
+# Publish PyPI and NPM artifacts
+name: Publish
+on:
+  push:
+  pull_request:
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: Build dist
+        run: |
+          pip install jupyter_packaging wheel jupyterlab~=3.0
+          python setup.py sdist bdist_wheel
+
+      - name: Javascript package
+        run: |
+          mkdir jsdist
+          cd jupyterlab-server-proxy
+          jlpm pack --filename ../jsdist/jupyterlab-server-proxy-jlpmpack.tgz
+
+      - name: Upload Python artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: dist
+          path: dist
+          if-no-files-found: error
+
+      - name: Upload Javascript artifact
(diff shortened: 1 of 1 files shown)


manics:
@consideRatio @yuvipanda Do you have access to https://pypi.org/project/jupyter-server-proxy/ and https://www.npmjs.com/package/@jupyterlab/server-proxy ? If so please could you add some secrets to this repo so we can release https://github.com/jupyterhub/jupyter-server-proxy/pull/245 when it's ready?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          password: ${{ secrets.PYPI_PASSWORD }}
+
+  # https://docs.github.com/en/actions/language-and-framework-guides/publishing-nodejs-packages#publishing-packages-to-the-npm-registry
+  publish-npm:
+    needs:
+      - build
+    runs-on: ubuntu-latest
+    steps:
+      # Setup .npmrc file to publish to npm
```
consideRatio, line 67:
```suggestion
      # actions/setup-node creates an .npmrc file that references NODE_AUTH_TOKEN
```

consideRatio:
@manics done!

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,81 @@
+# Publish PyPI and NPM artifacts
+name: Publish
+on:
+  push:
+  pull_request:
```
consideRatio, line 5: Providing a manual run button for those with push rights, something I've found generally useful for example to verify that something isn't broken on master that doesn't work in a PR I'm reviewing.

```suggestion
  pull_request:
  workflow_dispatch:
```

consideRatio (approved):
LGTM! Nice use of upload / download artifacts! :tada:

PR #249, opened 2021-02-18
Fix build.yaml workflow

manics:
It's a complete mystery to me why this didn't cause a CI failure until after the merge ....


consideRatio:
I think the reason is that YAML interpreted the script as a long string part of the name due to indentation :§

PR #252, opened 2021-02-25
Allow request header overrides

ryanlovett:
Add support for configurable HTTP headers.

### DIFF
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -138,7 +138,7 @@ pairs.
    ``--www-frame-origin=same`` to allow rendering of its pages in an iframe.
 
 
-``proxy_headers``
+``request_headers_override``
 ^^^^^^^^^^^^^^^
 
    One of:
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -138,6 +138,19 @@ pairs.
    ``--www-frame-origin=same`` to allow rendering of its pages in an iframe.
 
 
+``proxy_headers``
+^^^^^^^^^^^^^^^
+
+   One of:
+
+   * A dictionary of strings that are passed in as HTTP headers to the proxy
+     request. The strings ``{port}`` and ``{base_url}`` will be replaced as
+     for **command**.
+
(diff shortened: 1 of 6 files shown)


ryanlovett:
@manics I created this branch back in Feb., then rebased to fixup my own commits, then rebased onto the end of master. I hope I didn't mess anything up.

I created this feature to [enable support for X-RStudio-Root-Path](https://github.com/rstudio/rstudio/issues/8888#issuecomment-784517078) in the rsession proxy, though an early test showed it didn't help with that issue. Still, the feature may be useful for others.

yuvipanda:
Hey Ryan! Thanks for setting this up :)

So I'm guessing these are just for *request headers* - going to the backend from the proxy. If the header is already present, it is just overwritten, right?

In that case, I'd suggest:

1. Call it `request_headers_override` or something similar.
2. This dictionary will be merged with actual request headers and be sent to the backend. Headers set in `request_headers_override` will be preferred over any requests actually in the request from the user
3. We can add equivalents for `response_headers_override` in the future if needed.

For the callable, we could possibly pass in the current set of headers as an arg? Not sure how to go about that. What do you think?

ryanlovett:
@yuvipanda I agree that "request" is needed for disambiguation, thanks! What do you think about `custom_request_headers` or `updated_request_headers`? I think "override" suggests it may disclude new headers. "custom" is more conventional but doesn't imply that existing headers could be overridden, while "updated" is less conventional, but mirrors the python dict method.

Let me get back to you on the callable.

yuvipanda:
Hmm, override to me is 'override the *list* of headers', rather than overriding a particular header. IMO, it's common enough we can get away with it.

ryanlovett:
@yuvipanda I've updated the attribute to `request_headers_override`. Regarding the callable, I can see how knowing what an existing header is might be useful if needing to alter it, but do you think setting that up could be left to a later PR?

yuvipanda:
Thanks a lot, @ryanlovett!

PR #255, opened 2021-02-28
Move build.yaml into test.yaml

manics:
Most of it is duplicated.

Note since build.yaml used `python -m pip install jupyterlab` this was equivalent to installing JupyterLab 3. This means we don't have any tests for JupyterLab 2.


consideRatio (approved):
LGTM, with the TODO in place I consider this ready for merge.

Issue #205, opened 2020-06-24
Consider renaming whitelist


jacobtomlinson:
Should we consider renaming the `whitelist` option to `allowlist`?

manics:
👍 to renaming, related discussion here: https://github.com/jupyterhub/team-compass/issues/299

PR #256, opened 2021-03-07
Breaking: Replace host_whitelist with host_allowlist

manics:
Replaces `ServerProxy.host_whitelist` with `host_allowlist`, outputs a deprecation message if the old property is set.

Closes https://github.com/jupyterhub/jupyter-server-proxy/issues/205
This is based on https://github.com/jupyterhub/jupyterhub/pull/3090

This may not work in subclasses of `ServerProxy`, so best to treat this as a breaking change- since we've got a pending 3.0.0 release this seems like a good time to sort this out.

~This currently fails since `ServerProxy` is a Configurable which does not have a `log` attribute, only Applications do: https://traitlets.readthedocs.io/en/stable/config.html
Do you think I should just use the standard `logging.warn` instead even though it means the Application's logging config is ignored? Or do you have another suggestion?~


consideRatio (approved):
This LGTM and I agree this should be part of the 3.0.0 release!

minrk:
> This may not work in subclasses of ServerProxy

The only case that should be broken for subclasses is if they *refer* to the deprecated name, which may not be up-to-date if the updated config is used. I suspect that's rare, if ever.

FWIW, the pattern used in juptyerhub would make it `allowed_hosts` instead of `host_allowlist` but this works, too!

PR #259, opened 2021-03-16
Fix PyPI url

### LINKED ISSUE #258: PyPI release URL has changed
Following up on https://github.com/conda-forge/jupyter-server-proxy-feedstock/pull/20, we've noticed that the PyPI release url has changed.

Instead of `https://pypi.io/packages/source/j/jupyter-server-proxy/jupyter-server-proxy-3.0.0.tar.gz` it is now `https://pypi.io/packages/source/j/jupyter-server-proxy/jupyter_server_proxy-3.0.0.tar.gz`.

I believe this has to do with https://github.com/jupyterhub/jupyter-server-proxy/blob/master/setup.py#L20.

janjagusch:
Closes #258 

```sh
$ ls dist
jupyter-server-proxy-3.0.0.tar.gz           jupyter_server_proxy-3.0.0-py3-none-any.whl
```

The `.tar.gz` file now has the correct name.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ long_description = (HERE / "README.md").read_text()
 pkg_json = json.loads((HERE / "jupyterlab-server-proxy" / "package.json").read_bytes())
 
 setup_args = dict(
-    name=name,
+    name=name.replace("_", "-"),
     version=pkg_json["version"],
     url=pkg_json["homepage"],
     author=pkg_json["author"]["name"],


janjagusch:
@consideRatio, this should solve our issue in https://github.com/conda-forge/jupyter-server-proxy-feedstock/pull/20 as well. :)

setup.py
```diff
@@ -73,7 +73,7 @@
 pkg_json = json.loads((HERE / "jupyterlab-server-proxy" / "package.json").read_bytes())
 
 setup_args = dict(
-    name=name,
+    name=name.replace("_", "-"),
```
consideRatio, line 76: There are other references to the `name`, do you think it is correct for those to not be changes alongside this?

janjagusch: I believe the other `name` references should keep the `_` because we're actually referencing a directory in those cases. The entire npm build will break if we rename the original `name` variable (I tried that first, actually).

By only changing it in the `setup_args` I think the entire build process stays the same and the only thing that changes is the name of the `.tar.gz` file.

consideRatio:
@janjagusch nice I've released 3.0.1 now with this!

PR #263, opened 2021-03-27
Reduce (and test) sdist size

bollwyvl:
The 3.0.2 sdist clocks in at 43.3M.

This PR explicitly removes the `node_modules` from the sdist by updating MANIFEST.in, bringing the `.tar.gz` down to 82k, and adds a check in the publish CI (with a >2x margin for future growth).

To maintain reproducibility, it removes `yarn.lock` from `.gitignore`, which despite the commit thrash, is actually quite useful in tracking down where something might have changed, and will also save re-solving in CI.


bollwyvl:
Looks like having `yarn.lock` cuts about ~30s off the build. It can be more impactful to do one of the builds first, then run the second with `--install-option="--skip-npm"`, which avoids the re-run of the build machinery. Happy to do so on this PR, but otherwise nothing planned here.

.github/workflows/test.yaml
```diff
@@ -55,6 +55,10 @@ jobs:
         run: |
           pip install jupyter_packaging wheel jupyterlab
           python setup.py sdist bdist_wheel
+      - name: Check Package Sizes
+        run: |
+          ls -lathr dist
+          find dist -type f -size +200k | grep dist || echo ok
```
manics, line 61: Is this just for info, or should it fail if there's a file > 200k ? At the moment the exit code will always be 0 regardless.

manics (commented):
One comment, but looks good to me.

.github/workflows/test.yaml
bollwyvl: yeah, i need to figure out how to make it fail properly. will try some more stuff locally.

bollwyvl:
Sorry for the thrash (feel free to squash merge):
- 1437987  shows [the fail for 20k](https://github.com/jupyterhub/jupyter-server-proxy/pull/263/checks#step:5:16)
- 33b424c shows [the pass for 200k](https://github.com/jupyterhub/jupyter-server-proxy/pull/263/checks#step:5:16)

.github/workflows/test.yaml
bollwyvl: should be fully resolved by 33b424c

manics (approved):
Sorry for the delay, LGTM.

Is everyone else happy with the addition of `yarn.lock`? dependabot will update it, but they'll ber loads of PRs, e.g. see https://github.com/manics/jupyter-offlinenotebook/pulls?q=is%3Apr+is%3Aclosed+label%3Ajavascript

bollwyvl:
> ber loads of PRs, e.g. 

This hardly seems like a _bad_ thing... anyhow, any thoughts here, folks? That big ol' tarball is still out there burning bits-on-the-wire...

manics:
I guess silence is acquiescence? :smile: 

Could you help we with one question- what's the expectation when it comes to merging dependabot PRs to `yarn.lock`? Does the presence of pinned dependencies in this repo imply a guarantee to developers that they work? We don't have any JupyterLab UI tests so unless it breaks the build CI will always be green. Should we always do a manual UI test of the JupyterLab extension, or is it OK to just merge and assume it's fine?

bollwyvl:
The cheapest acceptance test is dev-focused binder, which seems useful _anyway_.

If you want full acceptance testing, I'm happy to spin up some robotframework-jupyterlibrary which can test classic and lab (and watch the server logs) and generate nice reports.

If this is a blocking concern, I can take out the yarn.lock, and have even _less_ assurance that the extension works.

manics:
I've added https://github.com/jupyterhub/jupyter-server-proxy/issues/268
If you've got time to add even some basic UI tests that would be great! I've been waiting for https://github.com/jupyterlab/extension-examples/issues/45 to find out how to do it properly.

Issue #223, opened 2020-11-09
Let port config option be a callback


LucasVerneyDGE:
Hi,

First of all, thanks for this very useful Jupyter(Lab) add-on!

### Proposed change

It would be really helpful if the port config option could be a callback, just the same as what is supported for the command.

### Alternative options

An alternative option would be to have per-user settings in the global configuration.

### Who would use this feature?

I have a specific set up where I am using jupyter-server-proxy to spin up IDE (such as Theia) for users. The easiest way to do it is of course to use the default attribution of port (which is great!) but this means any user can connect to another user's IDE provided they know the port (which they can find using Jupyter shell).

Therefore, as I am managing a limited number of users I thought about resorting to static port allocation by user (saying userA has portA for IDE, userB has portB etc.) and using a firewall rule to prevent cross-user traffic. This works great by putting the proxy configuration in each user local configuration (`$HOME/.jupyter`).

The issue is that this requires allowing per-user configuration (which I'd like to be able to disable). This could be handled in the global configuration using something like

```python
import os

PORTS = {
   'userA': 99999,
   'userB': 99998,
   ...
}
DEFAULT_PORT = _get_random_port()

def _cmd_callback():
    port = _port_callback()
    return ['some-command', '--port=' + str(port)]

def _port_callback():
    return PORTS.get(os.environ.get('USER'), DEFAULT_PORT)

server_config = {
    'command': _cmd_callback,
    'port': _port_callback,
}
```

Would you accept a PR offering such a mechanism?

Thanks in advance,
Best,

PR #265, opened 2021-04-07
Initialize icon_path when entry disabled

guimou:
Hi!
I am using the [jupyter-lmod](https://github.com/cmd-ntrf/jupyter-lmod) extension to dynamically load modules inside a Jupyter container.
The module is able to show/hide icons depending on the availability of the module/executable. For example, this prevents people to try launching RStudio, because they see the icon, if it's not loaded. In this mechanism, the launcher entry has to be set to "disabled" at the beginning (as the module is not loaded). The module takes care of dynamically adding the launcher entry.
However, with the current code, the icon path is not initialized when the module is not enabled from the beginning. And as it's impossible to add more handlers on the same path afterwards (unless I'm mistaken but I tried it multiple ways), the icon can never be displayed.
Proposed modification is to always initialize the icon_path when it's present, even if the entry is disabled. It does not cost much in the end, even if this icon/path is never used. And if the launcher entry is there, even if disabled, it must be for a good reason, so why not initialize the icon_path?

### DIFF
--- a/jupyter_server_proxy/__init__.py
+++ b/jupyter_server_proxy/__init__.py
@@ -33,7 +33,7 @@ def _load_jupyter_server_extension(nbapp):
 
     icons = {}
     for sp in server_processes:
-        if sp.launcher_entry.enabled and sp.launcher_entry.icon_path:
+        if sp.launcher_entry.icon_path:
             icons[sp.name] = sp.launcher_entry.icon_path
 
     nbapp.web_app.add_handlers('.*', [


yuvipanda:
Thank you for opening this PR, @guimou. Will this still display the entries in the JupyterLab launcher, even if they are disabled?

guimou:
@yuvipanda Do you need any other info?

guimou:
@yuvipanda Hi! I'm bumping this because we are thinking/working on integrating [the extension I mentioned](https://github.com/cmd-ntrf/jupyter-lmod) into the [Elyra project](https://elyra.readthedocs.io/). Before proceeding I have some cleanup to do. And depending on the result of the PR, I will either just use the jupyter-server-proxy package, or proceed as I do currently which is forking/copying the code directly in my extension (which is not really efficient...).
So, is anything else needed on my part? Anyway, no bad feelings if you decide not to accept this change, I just need to know.
Thanks a lot!

consideRatio:
I'm cautious about merging this PR as it seems like a breaking change that is a workaround to a situation that can be resolved in non-breaking ways I think.

We have `enabled` as a toggle to mean something very specific, and this change seems to bypass that. If you want to enable/disable based on some just-in-time information, i think it would be more reasonable to allow for the `launcher_entry` to be a callable.

Would having it be a callable resolve your issue @guimou?

https://github.com/jupyterhub/jupyter-server-proxy/blob/ade470cbdb33563ce3e7669f7e35b99e5f7d23f9/docs/source/server-process.rst#L106-L122

## Related

- #223

consideRatio:
@guimou to help me as a maintainer and other maintainers focus our attention, I'll make a call here and close this PR for now. With that said, I'm personally very open to the idea of allowing launcher_entry to be a callable function and if you wish you can open a PR about it directly I think. Such PR would also need some documentation updates!

If a callable function don't meet your needs I would like to ask that you open an issue to discuss what could meet your needs and how/if that could be reasonable to do.

guimou:
Hi @consideRatio !
No worries, it was ok closing this. Tbh I did not really think about it any more. For the moment I'm patching the installation  in my deployments. As it's very specific that's OK. I will look into the callable function option when I have time. It sounds good, but the constraint we have here, unless I'm mistaken, is that all of this must happen when Tornado launches. After it's initialized and loaded you cannot modify this type of resource (at least that's what I understood).
Thanks anyway for everything you and the other maintainers are doing!

PR #267, opened 2021-04-18
enable multiple launchers (with paths)

### LINKED ISSUE #222: Pass URL filename and URL parameters to browser
Some webservices can be initialized directly by passing parameters in the URL - for example this is the case for Xpra:  
`http://localhost:10000/index.html?username=test&password=123456`

If one wants to use jupyter-server-proxy to start a xpra html5 desktop one would first start xpra on the server, generate a one-time-password and might think it would be fine to add this password using a `mappath` command as the password uri parameters to `/index.html`

But this is will not work (!),  
because the `index.html` of Xpra wants to reads the password through the url parameters with javascript on the client side, too.

### Proposed change
Url parameters should be passed to the browser.

### Alternative options
The url parameters are shown to the user and he has to add it manually to the URL in the browsers address line.

Any webservice, which extracts information from the url parameter (using JavaScript) on client side.

### LINKED ISSUE #266: Allow redirecting to something(s) (other than /)
### Proposed change

Allow a single server instance to open different URLs with different labels, icons, etc. Even if a single path is given, it can be nice to pre-seed it with a specific path, including `?` params and `#` fragments (#222).

### Alternative options

Using multiple servers is unsatisfying if the route in question offers a number of different endpoints, each of which might be relevant, or in the case of Lab, might be interesting to have open simultaneously.

Some applications aren't particularly well-behaved (especially when run under a proxy) and need a little help to know where they are.

- the dask dashboard has a number of different views that can be launched
- [jupyterlite](https://github.com/jtpio/jupyterlite) offers a number of applications (lab, classic) that can "talk" to each other and it would be nice to have a single server to run them

In config, allow `launcher_entry` to be an array. Probably also allow `new_browser_tab` to be set per-launcher.

In the frontsend, a multi-launcher as a menu could have submenus:

```
JupyterLite >
  Lab
  Classic
  Book
Dask >
  Task Stream
  Profiler
  ...
```

Meanwhile, in the Lab launcher would draw multiple cards, command palette entries, etc.

bollwyvl:
- fixes #266
- fixes #222

This reworks the configuration in a backwards-compatible way to enable multiple launchers from the same server process.

The public REST API would change (from `launcher_entry` to `launcher_entries`), as does the URL scheme for the icons `<server.name>/<launcher.name>`.

I opted not to do the submenus, but could.

Additionally, in lab, this makes launcher and layoutrestorer optional, and adds mainmenu and commandpalette, which _should_ make this compatible with jupyterlab-classic (though it wouldn't be exposed in the son-of-tree page, and is probably out of scope for this PR).

I didn't see where launcher_entry is under test, but am happy to add something if requested.

### DIFF
--- a/jupyterlab-server-proxy/package.json
+++ b/jupyterlab-server-proxy/package.json
@@ -41,18 +41,20 @@
     "install:extension": "jupyter labextension develop --overwrite .",
     "prepare": "jlpm run clean && jlpm run build:prod",
     "watch": "run-p watch:src watch:labextension",
-    "watch:src": "tsc -w",
+    "watch:src": "tsc -w --preserveWatchOutput",
     "watch:labextension": "jupyter labextension watch ."
   },
   "dependencies": {
     "@jupyterlab/application": "^2.0.0 || ^3.0.0",
     "@jupyterlab/apputils": "^2.0.0 || ^3.0.0",
-    "@jupyterlab/launcher": "^2.0.0 || ^3.0.0"
+    "@jupyterlab/launcher": "^2.0.0 || ^3.0.0",
+    "@jupyterlab/mainmenu": "^2.0.0 || ^3.0.0"
   },
   "devDependencies": {
     "@jupyterlab/builder": "^3.0.2",
-    "rimraf": "^2.6.1",
-    "typescript": "~3.7.0"
+    "rimraf": "^3.0.2",
+    "typescript": "~4.2.4",
+    "npm-run-all": "~4.1.5"
   },
   "jupyterlab": {
     "extension": true,
--- a/jupyter_server_proxy/__init__.py
+++ b/jupyter_server_proxy/__init__.py
@@ -34,8 +34,10 @@ def _load_jupyter_server_extension(nbapp):
     launcher_entries = []
     icons = {}
     for sp in server_processes:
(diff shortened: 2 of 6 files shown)


bollwyvl:
ha, bash ate my commit message, but the most recent push should be:
```
add docs for `name` and `path`
```

jupyterlab-server-proxy/package.json
```diff
@@ -41,18 +41,20 @@
...
+    "@jupyterlab/launcher": "^2.0.0 || ^3.0.0",
+    "@jupyterlab/mainmenu": "^2.0.0 || ^3.0.0"
   },
   "devDependencies": {
     "@jupyterlab/builder": "^3.0.2",
-    "rimraf": "^2.6.1",
-    "typescript": "~3.7.0"
+    "rimraf": "^3.0.2",
+    "typescript": "~4.2.4",
+    "npm-run-all": "~4.1.5"
```
bollwyvl, line 57: makes `watch` work as advertised. the typescript/rimraf stuff is just a ride-along.

yuvipanda:
Better nate than lever. Sorry for the late response in reviewing this, @bollwyvl. Wanna rebase and we can try land this?

PR #271, opened 2021-04-30
More precise regexp matching for /proxy/absolute/<host>:<port>

### LINKED ISSUE #270: Failure to match `/proxy/absolute/<host>:<port>/`
### Bug description
`/proxy/absolute/<host>:<port>` does not work as intended, due to over-greedy regexp match which thinks that `absolute/` is part of the hostname.

Patch including test to follow.

#### Expected behaviour
Proxy to given host and port with unchanged path

#### Actual behaviour
Error is shown:

```
Host 'absolute/10.12.253.4' is not allowed. See https://jupyter-server-proxy.readthedocs.io/en/latest/arbitrary-ports-hosts.html for info.
```

### How to reproduce
1. Add `10.12.253.4` to host_allowlist
2. Access `/proxy/absolute/10.12.253.4:3080/v2/compute/`
3. Observe error above

### Your personal set up

 - OS: macOS 10.14.6
 - Version(s): jupyterlab 3.0.14, conda 4.10.1 / python 3.8.5.final.0

candlerb:
Fixes #270

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -568,9 +568,9 @@ class SuperviseAndProxyHandler(LocalProxyHandler):
 def setup_handlers(web_app, host_allowlist):
     host_pattern = '.*$'
     web_app.add_handlers('.*', [
-        (url_path_join(web_app.settings['base_url'], r'/proxy/(.*):(\d+)(.*)'),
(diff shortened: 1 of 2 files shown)


manics (approved):

manics:
Thanks for the fix!

PR #273, opened 2021-05-05
Open browser not required for running pytests

candlerb:
This is just a minor documentation tweak.  Starting the server to run tests is much faster if you don't open a browser window as well.


yuvipanda:
Thank you very much, @candlerb! This *does* make testing faster by quite a bit.

PR #278, opened 2021-06-25
Feature/add launcher url

zeehio:
This commit allows a provider of an proxied server to specify where the link in the launcher will point to.

Usually (and by default) this will be `/`, so the launcher can point to `{base_url} / {server_name} /`, however in some cases the entrypoint url to the launched service can be different (e.g.  `{base_url} / {server_name} / auth-sign-in` ...)

### DIFF
--- a/jupyter_server_proxy/api.py
+++ b/jupyter_server_proxy/api.py
@@ -20,7 +20,7 @@ class ServersInfoHandler(JupyterHandler):
                 'launcher_entry': {
                     'enabled': sp.launcher_entry.enabled,
                     'title': sp.launcher_entry.title,
-                    'launcher_url': sp.launcher_entry.launcher.url
+                    'launcher_url': sp.launcher_entry.launcher_url
                 },
                 'new_browser_tab' : sp.new_browser_tab
             }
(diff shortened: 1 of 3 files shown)


zeehio:
I don't have the technical skills to make changes to the jupyterlab launcher icons so they honor the `launcher_url` defined here.

However from the jupyter notebook menu it works.

ryanlovett:
Thanks @zeehio! This is very similar to my [path_info branch](https://github.com/jupyterhub/jupyter-server-proxy/compare/master...ryanlovett:path_info?expand=1), so great minds think alike. :) I was also targeting the rsession proxy issue. Like you I didn't feel comfortable with the lab launcher bits which is why I didn't get much further. At this point I wouldn't be averse to merging without that for now.

My only suggestion would be to name the variable something other than launcher_url, since that sounds like it sets the whole URL rather than appends to the existing one. (even though you documented it) What do you think?

zeehio:
I saw yet another pull request (I forgot which one) with something similar to ours as well.

Your name seems better than mine to me. Can you merge your pull request? I will happily close mine.

Looking forward to fixing jupyter-rsession-proxy!

riazarbi:
I've conducted an informal test of this MR here - 

https://github.com/jupyterhub/jupyter-rsession-proxy/pull/103#issuecomment-868977588

zeehio:
See #279 for a similar merged solution! :-)

jhgoebbert:
Hi @zeehio, thank you for your path_info.

I wonder if you have tested path_info with JupyterLab. To me it looks as if it is missing in 
https://github.com/jupyterhub/jupyter-server-proxy/blob/master/jupyterlab-server-proxy/src/index.ts#L83

PR #279, opened 2021-07-03
Path info

ryanlovett:
Similar to @zeehio's PR #278, this enables proxied servers to alter the launcher's path. By default it will remain the name of the proxied server.

This will enable jupyter-rsession-proxy to augment the launcher path, working around an rserver issue.


jhgoebbert:
This MR should fix it: https://github.com/jupyterhub/jupyter-server-proxy/pull/294

Issue #281, opened 2021-07-06
Connectio lost (HTTP ERROR 599) when downloading large files through proxy


pisymbol:
### Bug description
Can not download large files through minio behind proxy 

#### Expected behaviour
File should be downloaded

#### Actual behaviour
[E 2021-07-06 10:36:40.919 ServerApp] {
      "Host": "127.0.0.1:8888",
      "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:89.0) Gecko/20100101 Firefox/89.0",
      "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8",
      "Accept-Language": "en-US,en;q=0.5",
      "Accept-Encoding": "gzip, deflate",
      "Connection": "keep-alive",
      "Referer": "http://127.0.0.1:8888/minio/my-data/",
      "Cookie": "username-127-0-0-1-8888=\"2|1:0|10:1625567544|23:username-127-0-0-1-8888|44:NjE4MDZkY2ExNmQ0NGE1ZjljNzJiMjBjNTMxM2ZkNjc=|9c563251fbaace553f943e1b56085bb0a7f393aca4955c87e3f975004468eb27\"; _xsrf=2|a3b32fd7|d207e1bdf72df12803c565b723922e6d|1625153033",
      "Upgrade-Insecure-Requests": "1"
    }
[E 2021-07-06 10:36:40.919 ServerApp] 599 GET /minio/download/my-data/foo.deb?token=eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhY2Nlc3NLZXkiOiJtaW5pb2FkbWluIiwiZXhwIjoxNjI1NTY3ODYwLCJzdWIiOiJtaW5pb2FkbWluIn0.OlQeJhkVXCaYs3VLhDe4A00QqrRIzDsR0Lhl3AADbgOtRHyZsEARoyjixU6IDbzsPN0OAuxlxCfT4HA6Ekecvg (172.17.0.1) 4.97ms referer=http://127.0.0.1:8888/minio/my-data/

minio setup:

   'command': _get_minio_command,
        'port': 9000,
        'timeout': 120,
        'new_browser_tab': True,
        'absolute_url': True,
        'launcher_entry': {
            'title': 'MinIO',
            'icon_path': _get_icon_path()
        },

_get_minio_command() just returns 'minio server /data' etc.

[...]

PR #282, opened 2021-07-08
Add proxy_client_options

pisymbol:
Issue #281 Allow a per server process client configuration of the underlying HTTP
proxy client.

For example, the current web framework, Tornado, limits
the maximum body size to 100MiB which can restrict how much data
the proxy can request from the backend server process, e.g.
imagine proxying a storage interface like MinIO from within Jupyter.

Since this extension is not strictly limited to Tornado per se, the
"proxy_client_options" config is just a dictionary that gets passed in to
client as is.

Now with this changeset, the underlying httpclient.AsyncHTTPClient
can be configured like so:

```
'command': getsome_server_command(),
...
'proxy_client_options': {
	'max_body_size': 1024*1024*1024,
},
```

### DIFF
--- a/docs/server-process.rst
+++ b/docs/server-process.rst
@@ -150,6 +150,15 @@ pairs.
    * A callable that takes any :ref:`callable arguments <server-process/callable-argument>`,
      and returns a dictionary of strings that are used & treated same as above.
 
+``proxy_client_options``
+^^^^^^^^^^^^^^^
+
+   One of:
+
+   * A dictionary of arguments that are used to construct the proxy HTTP client.
+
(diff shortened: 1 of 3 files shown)


yuvipanda:
Apologies for such a late response, @pisymbol.

I would like us to not expose the internal details of the HTTP client we are using to end users like this, and instead expose specific options (such as request timeouts) as needed instead. This allows us the freedom to change HTTP libraries in the future if we so desire, without breaking everyone.

Apologies again for the late response, and the late response being a close. Happy to accept specific documented options you need that can be passed through, though.

PR #287, opened 2021-08-17
Apply `request_headers_override` to websocket requests

### LINKED ISSUE #286: Apply `request_headers_override` to websocket requests
### Proposed change
It would be really helpful if request headers to websocket endpoints could be overridden, too. I would suggest to just apply `request_headers_override` also to websocket requests. An example use-case is authentication using API key headers. This is only really workable if the API key header is applied to all requests, and that includes websocket connections.

### Alternative options
As an alternative, an API key could also be appended to the URL as a query parameter using `mappath`. But currently, this is not possible: if `mappath` returns a path including a query parameter, the original query parameter from the request is [only string-appended to the URI](https://github.com/jupyterhub/jupyter-server-proxy/blob/1e0cbb3816210388bdb64e78ef50aaccdc6c544b/jupyter_server_proxy/handlers.py#L163) - the path returned from `mappath` is assumed not to contain query parameters.

This feature is useful for jupyter deployments where multiple users are not isolated using network namespaces (i.e. containers), and the connection between the app and the proxy needs to be authentication to prevent privilege escalation, for applications that establish websocket connections.

sk1p:
Fixes #286

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -282,7 +282,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
             proxied_path = '/' + proxied_path
 
         client_uri = self.get_client_uri('ws', host, port, proxied_path)
-        headers = self.request.headers
+        headers = self.proxy_request_headers()
         current_loop = ioloop.IOLoop.current()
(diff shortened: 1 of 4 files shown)


yuvipanda:
Thank you very much for the PR and the detailed tests, @sk1p! I really appreciate the effort you've put into making this easy to review and merge

PR #290, opened 2021-08-27
keep gzip-encoded content compressed

### LINKED ISSUE #288: keep gzip-encoded content compressed
### Bug description
When serving gzip-encoded content through the proxy, the content is decompressed by the proxy.

#### Expected behaviour
When serving gzip-encoded content through the proxy:
- The content is kept compressed
- The header `Content-Encoding: gzip` is kept

#### Actual behaviour
When serving gzip-encoded content through the proxy:
- The content is decompressed by the proxy
- The header `Content-Encoding: gzip` is dropped
- The header `X-Consumed-Content-Encoding: gzip` is added instead

### How to reproduce
- Add `jupyter_notebook_config.py`:
```
c.ServerProxy.servers = {
  'gzipserver': {
    'command': ['python', '/path/to/gzipserver.py'],
    'port': 8080,
  }
}
```
- Add `gzipserver.py`:
```
import http.server
from io import BytesIO
import gzip

class GzipServer(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        fileobj = BytesIO()
        f = gzip.GzipFile(fileobj=fileobj, mode='w')
        f.write(b"this is a test")
        f.close()
        content = fileobj.getvalue()
        self.send_response(200)
        self.send_header("Content-length", str(len(content)))
        self.send_header("Content-type", "text/html")
        self.send_header("Content-Encoding", "gzip")
        self.end_headers()
        self.wfile.write(content)
        self.wfile.flush()

http.server.HTTPServer(('', 8080), GzipServer).serve_forever()
```
- Start jupyterlab
- Run:
```
curl -L -v http://127.0.0.1:8888/gzipserver
```
- Output ('`...`' means that parts of the output were removed):
```
...
< X-Consumed-Content-Encoding: gzip
...

this is a test
```

### Your personal set up
zero-to-jupyterhub on kubernetes using our own notebook image, but it can be reproduced using a standalone notebook image with `docker run`.

- OS: CentOS 7.9
```
$ cat /etc/centos-release
CentOS Linux release 7.9.2009 (Core)
```

- Version(s):
```
$ jupyterhub --version
1.1.0
$ python --version
Python 3.6.8
```

### Suggested fix

I suggest fixing this by passing `decompress_response=False` to the `HTTPRequest` constructor, and not filtering out the `Content-Encoding` header from the response. See the patch on commit https://github.com/jupyterhub/jupyter-server-proxy/commit/1e0cbb3816210388bdb64e78ef50aaccdc6c544b below:
```
diff --git a/jupyter_server_proxy/handlers.py b/jupyter_server_proxy/handlers.py
index 12d86cd..c5a3683 100644
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -180,6 +180,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
 
         req = httpclient.HTTPRequest(
             client_uri, method=self.request.method, body=body,
+            decompress_response=False,
             headers=headers, **self.proxy_request_options())
         return req
 
@@ -256,7 +257,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
 
             for header, v in response.headers.get_all():
                 if header not in ('Content-Length', 'Transfer-Encoding',
-                                  'Content-Encoding', 'Connection'):
+                                  'Connection'):
                     # some header appear multiple times, eg 'Set-Cookie'
                     self.add_header(header, v)
 
```

axelmartenssonmodelon:
Fixes #288

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -180,6 +180,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
 
         req = httpclient.HTTPRequest(
             client_uri, method=self.request.method, body=body,
+            decompress_response=False,
             headers=headers, **self.proxy_request_options())
         return req
 
@@ -256,7 +257,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
 
             for header, v in response.headers.get_all():
                 if header not in ('Content-Length', 'Transfer-Encoding',
-                                  'Content-Encoding', 'Connection'):
+                                  'Connection'):
                     # some header appear multiple times, eg 'Set-Cookie'
                     self.add_header(header, v)
 
--- a/tests/resources/gzipserver.py
+++ b/tests/resources/gzipserver.py
@@ -1,0 +1,25 @@
+from http.server import HTTPServer, BaseHTTPRequestHandler
+from io import BytesIO
+import sys
+import gzip
+
+class GzipServer(BaseHTTPRequestHandler):
+    def do_GET(self):
+        fileobj = BytesIO()
+        f = gzip.GzipFile(fileobj=fileobj, mode='w')
+        f.write(b"this is a test")
+        f.close()
+        content = fileobj.getvalue()
+        self.send_response(200)
+        self.send_header("Content-length", str(len(content)))
+        self.send_header("Content-type", "text/html")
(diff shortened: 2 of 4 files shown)


manics (approved):
Thanks!

PR #298, opened 2021-11-04
Test with lab

### LINKED ISSUE #297: Replace jupyter-notebook by jupyter-lab
### Bug description

[In CONTRIBUTING.md](https://github.com/jupyterhub/jupyter-server-proxy/blob/7d7eae940285b92cff98a66f64d6fa9b8fe5506a/CONTRIBUTING.md?plain=1#L23) it is suggested to use Jupyter Notebook for testing. I haven't touched Notebook for years, and I think it would make more sense to use `jupyter-lab` in place of `jupyter-notebook`.

I also noticed something similar in the CI [where Notebook is referenced](https://github.com/jupyterhub/jupyter-server-proxy/blob/7d7eae940285b92cff98a66f64d6fa9b8fe5506a/.github/workflows/test.yaml#L40). I didn't understand the `matrix.include` construction, but I found ["Including additional values into combinations"](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations) as a source from which it was apparently copy-pasted and adapted. (As evidence, note the [irrelevant comment about npm](https://github.com/jupyterhub/jupyter-server-proxy/blob/7d7eae940285b92cff98a66f64d6fa9b8fe5506a/.github/workflows/test.yaml#L42).) But the "Including additional values into combinations" section which I linked doesn't seem applicable in this case, rather the ["Including new combinations"](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-including-new-combinations) example seems relevant.

There may be other places where some cleanup is warranted, and I'm not exactly sure how one should clean up the CI. Thus I haven't written a PR.

#### Expected behaviour

The world moves on to JupyterLab, and Notebook is slowly forgotten.

#### Actual behaviour

Notebook seems to still crop up in a few random places.

maresb:
Closes #297

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -38,12 +38,6 @@ jobs:
         jupyterlab_version: [2, 3]
         python: [3.6, 3.7, 3.8, 3.9]
         jupyter_app: [notebook, lab]
-        include:
-          # includes a new variable of npm with a value of 2
-          # for the matrix leg matching the os and version
-          - jupyterlab_version: 3
-            python: 3.9
-            jupyter_app: server
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,7 +37,7 @@ jobs:
         # gain meaning on how job steps use them.
         jupyterlab_version: [2, 3]
         python: [3.6, 3.7, 3.8, 3.9]
-        jupyter_app: [notebook]
+        jupyter_app: [notebook, lab]
         include:
           # includes a new variable of npm with a value of 2
           # for the matrix leg matching the os and version
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,11 +16,10 @@ pip install -e .[test]
 jupyter serverextension enable --sys-prefix jupyter_server_proxy
 ```
 
-Before running tests, you need a [notebook
-server](https://github.com/jupyter/notebook) that we can test against.
+Before running tests, you need a server that we can test against.
 
 ```
-JUPYTER_TOKEN=secret jupyter-notebook --config=./tests/resources/jupyter_server_config.py --no-browser
+JUPYTER_TOKEN=secret jupyter-lab --config=./tests/resources/jupyter_server_config.py --no-browser
 ```
 
 Run the tests:


consideRatio:
This seem to make sense to me. I think it is a quite intense test quite at this point, but I vote for a merge.

consideRatio (approved):

.github/workflows/test.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
         # gain meaning on how job steps use them.
         jupyterlab_version: [2, 3]
         python: [3.6, 3.7, 3.8, 3.9]
-        jupyter_app: [notebook]
+        jupyter_app: [notebook, lab]
```
manics, line 40: Aren't `server` and `lab` effectively the same? If so we can get of the `include:` below.

maresb: @manics, under the assumption that you're correct I have removed the extra test case for `server`.

bollwyvl: Yep, `server` on its own doesn't really do much, `jupyter lab` is what the docs would say.

CONTRIBUTING.md
```diff
@@ -16,11 +16,10 @@ pip install -e .[test]
 jupyter serverextension enable --sys-prefix jupyter_server_proxy
 ```
 
-Before running tests, you need a [notebook
-server](https://github.com/jupyter/notebook) that we can test against.
+Before running tests, you need a server that we can test against.
 
 ```
-JUPYTER_TOKEN=secret jupyter-notebook --config=./tests/resources/jupyter_server_config.py --no-browser
+JUPYTER_TOKEN=secret jupyter-lab --config=./tests/resources/jupyter_server_config.py --no-browser
```
bollwyvl, line 22: Instead of this rather convoluted invocation  (which wouldn't work on e.g. windows): we could hoist the server to a (parametrized) fixture which started either the `notebook` or `lab` process, depending on availability, and by default tested both.

maresb: That could be quite convenient. Maybe then it would be possible to coordinate which `notebook`/`lab` port to use. (Often I would try to run tests and port 8888 would be already in use.)

bollwyvl: haha, yeah, ports. On `robotframework-jupyterlibrary`, we actually [randomize the port](https://github.com/robots-from-jupyter/robotframework-jupyterlibrary/blob/master/src/JupyterLibrary/keywords/server.py#L245) and and the token every time.

Another thing, _particularly_ to which effective use of `jupyter-server-proxy` is _particularly_ sensitive, is `base_url`. It might be worth adding a "nasty" one to the matrix.

Once "parametrize all the tests" goes a bit too far, though, `hypothesis` can be super handy... but one typically wouldn't want to restart process (much less the browser) if it's going to run a couple thousand times. But i feel like `jupyter-(lab|notebook)` and `--base-url="(/|/a/very long/prefix/💩/)"` may well be warranted.

maresb: This sounds like a good idea. Would it make sense to open a separate issue for it? (Unfortunately I don't have any time to work on such stuff at the moment.)

bollwyvl: issue up! #309

manics (approved):
Thanks all!

PR #299, opened 2021-11-05
fix: do not follow redirects when checking if server is up

ableuler:
When simply checking if the server that we're proxying to is up, I think it would make sense to not follow redirects and just interpret redirection as success. In my case this fixed an infinite redirection loop which occurred when proxying to an Rstudio server.


manics (approved):
👍Thanks, this looks good to me.

It shouldn't break existing usage since as the comment says
> We only care if we get back *any* response, not just 200
> If there's an error response, that can be shown directly to the user

PR #300, opened 2021-11-06
Enable rewrite_response to modify status and headers

ryanlovett:
This builds on @maresb 's work in #209 and enables one to also alter response HTTP headers and status code/reason.  Thanks to @yuvipanda for the idea!

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -46,7 +46,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
         self.host_allowlist = kwargs.pop('host_allowlist', ['localhost', '127.0.0.1'])
         self.rewrite_response = kwargs.pop(
             'rewrite_response',
-            lambda host, port, path, response: {}
+            lambda request, host, port, path, response: {}
         )
         self.subprotocols = None
         super().__init__(*args, **kwargs)
@@ -257,10 +257,10 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
         else:
             # self.rewrite_response returns a dict of 'code', 'reason',
             # 'headers', and 'body'. The function definition is
-            #   lambda host, port, path, response: {}
+            #   lambda request, host, base_url, port, path, response: {}
             # unless overridden in configuration.
             rewritten_response = self.rewrite_response(
-                host, port, proxied_path, response
+                self.request, host, port, proxied_path, response
             )
 
             ## status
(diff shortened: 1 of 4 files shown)


tests/resources/jupyter_server_config.py
```diff
@@ -40,10 +39,14 @@ def mappathf(path):
     'python-gzipserver': {
         'command': ['python3', './tests/resources/gzipserver.py', '{port}'],
     },
+    'python-http-rewrite-response': {
```
yuvipanda, line 42: can you add tests here for code, headers, etc as well?

yuvipanda (approved):
Let's add a test, and then we're good to go!

jupyter_server_proxy/handlers.py
```diff
@@ -46,7 +46,7 @@ def __init__(self, *args, **kwargs):
         self.host_allowlist = kwargs.pop('host_allowlist', ['localhost', '127.0.0.1'])
         self.rewrite_response = kwargs.pop(
             'rewrite_response',
-            lambda host, port, path, response: response.body
+            lambda request, host, port, path, response: {}
```
maresb, line 49: It makes me really nervous to see a [mutable object inside a default argument](https://www.google.com/search?q=python+mutable+default+argument). But in this case I think it's effectively immutable thanks to the `lambda`.

maresb:
I really like the extra functionality that this adds, thanks @ryanlovett!

I feel wary about this dictionary and the large number of function arguments flying around, and I wonder if we can do better...

What if we create a `RewritableResponse` object which is some sort of `dataclass` with properties `request`, `host`, `port`, `path`, `response`, `code`, `reason`, `headers`, and `body`. Then we can make the `rewrite_response` function have signature `RewritableResponse -> RewritableResponse`.

In addition to making the code simpler, this would also make it easy to compose rewrite functions or use them as decorators.

yuvipanda:
While thinking through @maresb's point about unweily arguments, I realized what we should do is use https://github.com/jupyterhub/jupyter-server-proxy/blob/5fd3969efb31eae3b76a64860cf0de663ee55a92/jupyter_server_proxy/utils.py#L4. We use that [elsewhere](https://github.com/jupyterhub/jupyter-server-proxy/search?q=call_with_asked_args) to allow us to add arbitrary args in the future without breaking existing code. So users can request as few or as many args as they want, and we don't need to add extra layers of indirection (like an additional class).

maresb:
Hmm, I'm quite fond of the ability to chain rewrites in my implementation. Do you see any way to achieve this with `call_with_asked_args`? It seems to me like it would be much harder that way, but maybe I just haven't thought enough in that direction.

yuvipanda:
@maresb can you tell me a bit more about what you mean by 'chained rewrites'? An example would be lovely.

maresb:
@yuvipanda, answered in #301.

PR #301, opened 2021-11-06
Class-based rewrite response

maresb:
Building on #300, illustrating my idea of implementing a `RewritableResponse` class. I tried to use Traitlets to make my data class, since it's used throughout the Jupyter ecosystem, but I don't know it very well.

I was originally thinking of having `RewritableResponse` be the *return* type as well as the input type, but I found this awkward. Now I have `rewrite_response` mutating the `RewritableResponse` object. ~~But this makes the variable names a bit awkward... in order for the variable names to make sense, I now define `original_response` run `rewrite_response(original_response)`, define `rewritten_response` as `original_response` and then delete `original_response`. There must be a better way which I'm not seeing at the moment.~~

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -209,20 +209,22 @@ class ServerProxy(Configurable):
 
           rewrite_response
             An optional function to rewrite the response for the given service.
-            Input is a RewritableResponse object. The function should modify one or
-            more of the attributes ``.body``, ``.headers``, ``.code``, or ``.reason``.
-            For example:
+            Input is a RewritableResponse object which is an argument that MUST be named
+            ``response``. The function should modify one or more of the attributes
+            ``.body``, ``.headers``, ``.code``, or ``.reason`` of the ``response``
+            argument. For example:
 
                 def cat_to_dog(response):
                     response.body = response.body.replace(b'cat', b'dog')
 
                 c.ServerProxy.servers['my_server']['rewrite_response'] = cat_to_dog
 
-            The RewritableRespone object also has attributes ``.host``, ``.port``, and
-            ``.path`` corresponding to the URL ``/proxy/<host>:<port><path>``. In
+            The ``rewrite_response`` function can also accept several optional
+            positional arguments. Arguments named ``host``, ``port``, and ``path`` will
+            receive values corresponding to the URL ``/proxy/<host>:<port><path>``. In
             addition, the original Tornado ``HTTPRequest`` and ``HTTPResponse`` objects
-            are available as ``.raw_request`` and ``.raw_response``. (These attributes
-            should not be modified.)
+            are available as arguments named ``orig_request`` and ``orig_response``.
+            (These objects should not be modified.)
 
             A list or tuple of functions can also be specified for chaining multiple
             rewrites.
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -136,7 +136,7 @@ def make_server_process(name, server_process_config, serverproxy_config):
         request_headers_override=server_process_config.get('request_headers_override', {}),
         rewrite_response=server_process_config.get(
             'rewrite_response',
-            lambda host, port, path, response: response.body
+            lambda host, port, path, response: {}
         ),
     )
 
@@ -212,22 +212,22 @@ class ServerProxy(Configurable):
             port ``port``, the ``path`` from the requested URL, and
             ``response`` is a `tornado.httpclient.HTTPResponse object
             <https://www.tornadoweb.org/en/stable/httpclient.html#response-objects>`.
-            Output is bytes.
-            Defaults to ``lambda host, port, path, response: response.body``.
+            Output is a dictionary with optional keys for "code", "reason", "headers", and "body". If a key is unspecified, the corresponding response value will not be altered. The value types match those in a `tornado.httpclient.HTTPResponse` object.
+            Defaults to ``lambda host, port, path, response: {}``.
         """,
         config=True
     )
 
     non_service_rewrite_response = Callable(
-        lambda host, port, path, response: response.body,
+        lambda host, port, path, response: {},
         help="""
         A function to rewrite the response for a non-service request, for
         example a request to ``/proxy/<host>:<port><path>``. Input arguments
         ``host``, ``port``, and ``path`` come from the requested URL, and
         "response" is a `tornado.httpclient.HTTPResponse object
         <https://www.tornadoweb.org/en/stable/httpclient.html#response-objects>`.
-        Output is bytes.
-        Defaults to ``lambda host, port, path, response: response.body``.
+        See the description for ``rewrite_response`` for more information.
+        Defaults to ``lambda host, port, path, response: {}``.
         """,
         config=True
     )
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -223,8 +223,8 @@ class ServerProxy(Configurable):
             positional arguments. Arguments named ``host``, ``port``, and ``path`` will
             receive values corresponding to the URL ``/proxy/<host>:<port><path>``. In
             addition, the original Tornado ``HTTPRequest`` and ``HTTPResponse`` objects
-            are available as arguments named ``orig_request`` and ``orig_response``.
-            (These objects should not be modified.)
+            are available as arguments named ``request`` and ``orig_response``. (These
+            objects should not be modified.)
 
             A list or tuple of functions can also be specified for chaining multiple
             rewrites.
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -4,8 +4,9 @@ Traitlets based configuration for jupyter_server_proxy
 from jupyter_server.utils import url_path_join as ujoin
 from traitlets import Dict, List, Union, default, observe
 from traitlets.config import Configurable
+from tornado import httpclient
 from warnings import warn
-from .handlers import SuperviseAndProxyHandler, AddSlashHandler
+from .handlers import SuperviseAndProxyHandler, AddSlashHandler, RewritableResponse
 import pkg_resources
 from collections import namedtuple
 from .utils import call_with_asked_args
@@ -136,7 +137,7 @@ def make_server_process(name, server_process_config, serverproxy_config):
         request_headers_override=server_process_config.get('request_headers_override', {}),
         rewrite_response=server_process_config.get(
             'rewrite_response',
-            lambda request, host, port, path, response: {}
+            lambda response: None,
         ),
     )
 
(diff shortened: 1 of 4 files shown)


maresb:
We could also support `rewrite_response` as string to be `exec`-ed so that as a shortcut one could say
```python
rewrite_response="response.body = response.body.replace(b'cat', b'dog')"
```
but then there are potential security issues which I don't know enough how to address. Is there any potential privilege escalation here? (If you can modify the config file, can you already execute arbitrary code as the user running the server?)

yuvipanda:
Thanks for working on this, @maresb! See my comment in https://github.com/jupyterhub/jupyter-server-proxy/pull/300#issuecomment-962638174 about a path forward.

yuvipanda:
Ah, I see - as the objects passed in are mutated, you can just pass in the same objects to a bunch of different functions, and they will be able to do their modifications as a chain. This lets you compose these changes - is that right?

maresb:
Exactly @yuvipanda.

I can do
```python
rewrite_response=[rewrite_function_a, rewrite_function_b]
```
and then they run one after the other.

maresb:
Hi @ryanlovett, please give [`raw_request`](https://github.com/jupyterhub/jupyter-server-proxy/blob/40593f3cdd1ce98748f6a70f5861a812e2cfc53a/jupyter_server_proxy/handlers.py#L58-L60)`.url` a try. I haven't tested it, but I think it should work in theory.

I still need to document things fully.

yuvipanda:
Paging @minrk to the API design discussion as well. I love the idea of chaining, but modifying arguments passed in implicitly makes me feel bad. I don't know how to balance these feelings! Help, @minrk!

Two things to think about:

1. If we modify passed in objects, do we still need a separate specialized class? I'm hoping we won't, and can instead just pass in the args the hook asks for. That's what we do for everyone else, right?
2. Is there a way to do chaining without having to modify the passed in arguments?

My intuition is that I feel stronger about (1) than (2).

Thank you for being awesome and pushing this forward, @moresb and @ryanlovett!

ryanlovett:
Ah, sorry I missed that @maresb . I'll give that a try.

minrk:
This looks great! Like @yuvipanda, my thought is that if it's mutating the object, it can be (a copy of) the original object, rather than a custom class?

> modifying arguments passed in implicitly makes me feel bad.

Modifying inputs for rewriting a response doesn't feel bad to me at all, so I have no problem with that. I don't think it's implicit when it's the sole purpose of the function.

If @yuvipanda feels strongly about immutability, then I think a custom class is appropriate, and each step in the chain would get a new instance of the class. I'm not sure what problem that solves, though.

maresb:
@minrk and @yuvipanda, thanks so much for the feedback.

I actually came to the same conclusion (that I could mutate a copy of the class). It took a little extra boilerplate (no `copy()` method in Traitlets!?) but my latest commit implements this.

minrk:
>  (no copy() method in Traitlets!?)

There's no copy method because copy is in the standard library and works for most objects: `obj_copy = copy.copy(obj)`, but your method works, too!

jupyter_server_proxy/config.py
```diff
@@ -208,26 +209,35 @@ class ServerProxy(Configurable):
 
           rewrite_response
             An optional function to rewrite the response for the given service.
-            Input arguments are ``host`` which is ``"localhost"``, the service
-            port ``port``, the ``path`` from the requested URL, and
-            ``response`` is a `tornado.httpclient.HTTPResponse object
-            <https://www.tornadoweb.org/en/stable/httpclient.html#response-objects>`.
-            Output is bytes.
-            Defaults to ``lambda host, port, path, response: response.body``.
+            Input is a RewritableResponse object. The function should modify one or
```
yuvipanda, line 212: Input should also be given the request object if it asks for it.

yuvipanda (changes requested):
One suggestion about passing in the request object too, and I think we're good to go after that!

Thanks a lot, @maresb, @minrk and @ryanlovett!

jupyter_server_proxy/config.py
maresb: @yuvipanda, does my latest commit properly address how to access the request object?

maresb:
I'm happy that you're convinced by my PR. Now that I know it has support, I think I should add a few quick tests before merging. (For example, to verify that chained rewrites work properly, and that the request object is accessible.)

jupyter_server_proxy/config.py
```diff
@@ -208,26 +209,38 @@ class ServerProxy(Configurable):
...
+            Defaults to the empty tuple ``tuple()``.
         """,
         config=True
     )
 
-    non_service_rewrite_response = Callable(
-        lambda host, port, path, response: response.body,
+    non_service_rewrite_response = Union(
+        default_value=tuple(),
+        trait_types=[List(), Tuple(), Callable()],
```
yuvipanda, line 237: Is there a reason we should support both Lists and Tuples?

jupyter_server_proxy/config.py
yuvipanda: Can we pass them as params when asked for instead? I am generally a big fan of not creating custom class hierarchies unless required :D

maresb: Hmm, I thought you and minrk just gave a thumbs-up to my custom `RewritableResponse` class. It is supposed to be a container for all the necessary arguments. If you don't like it then we have to start from scratch, right? Or maybe I'm misunderstanding what you're saying.

(I have a very strong distaste myself for excessive OOP, but here it seems to me like it serves a good purpose.)

yuvipanda: @maresb so my understanding is that there are *two things* - a `Response` that is coming back from our upstream server that we are rewriting (and hence the `RewritableResponse`, but also a readonly *request* that came from the client that triggered this response eventually. Right now, I think that request is inside this response object, even though from a pure HTTP perspective they aren't related. So in my mind, the function signature would be something like `def rewrite(orig_request: tornado.web.HTTPServerRequest, response: RewriteableResponse)` - hope that makes sense? So this shouldn't be starting from scratch as much as making `RewriteableResponse` a pure analogue of a hypothetical `Response` object. The original_request is readonly so shouldn't affect chaining.

maresb:
Hi @yuvipanda, @minrk and @ryanlovett,

My PR seems to have stalled out. I'm waiting for feedback on whether or not to add extra parameters (besides `RewritableResponse`) to the `rewrite_response()` function. Then I would like to write tests and hopefully get this merged.

I'll try and summarize the status here:

@yuvipanda wants the `rewrite_response` signature to be `rewrite_response(orig_request: tornado.httpclient.HTTPRequest, response: RewriteableResponse)`. I argue that this isn't necessary since as with `tornado.httpclient.HTTPResponse`, the request is provided as an attribute of `RewritableResponse`. I think it's more elegant to keep the signature short so that the `rewrite_response()` author only has to deal with the arguments they need. If you're unconvinced by my argument, I'm happy to change the signature as you suggest to move things forward.

Then there's @minrk's comment, for which I see two different and interesting interpretations, both of which I have already tried to address.

> my thought is that if it's mutating the object, it can be (a copy of) the **original** object

I originally interpreted "original" to mean the original where I interpret "original" to mean my original `RewritableResponse` object which I was awkwardly mutating. By implementing an `.apply_to_copy(rewrite_function)` method, I was able to eliminate my homegrown `.copy()` method and clean up my awkward code. :heavy_check_mark: 

Now I realize that @minrk likely indended that I could eliminate my `RewritableResponse` class entirely in favor of copying the `tornado.httpclient.HTTPResponse` object. This has some challenges which I don't yet see how to overcome. One annoyance is that many attributes are classes, so I don't know how "deep" I should make my copy. The more difficult challenge is that Tornado implements `.body` as a read-only property. I'd like to override this in the copied instance, but the descriptor lives in Tornado's `HTTPResponse` class, which I don't want to monkey patch. I could subclass Tornado's `HTTPResponse` and add a setter, but I don't know how to cast the copy to my subclass. (I actually tried by modifying `.__class__`, but that led to some very bizarre errors.)

Any suggestions on if / how to move this PR forward? Thanks!

maresb:
Sorry, I lost track of @yuvipanda's comment in the previous PR regarding `call_with_asked_args`. Now his request makes a lot more sense...

maresb:
Ok, if I get the green light from @yuvipanda regarding the implementation, I'll write some tests. Thanks for your patience!

jupyter_server_proxy/handlers.py
```diff
@@ -255,22 +297,62 @@ async def proxy(self, host, port, proxied_path):
...
+            rewrite_response = self.rewrite_response
+
+            # If this is a single function, wrap it in a list.
+            if isinstance(rewrite_response, (list, tuple)):
+                rewrite_responses = rewrite_response
+            else:
+                rewrite_responses = [rewrite_response]
+
+            # To be passed on-demand as args to the rewrite_response functions.
+            optional_args_to_rewrite_function = {
```
yuvipanda, line 314: Since these were never in a released version, I think just *request* is enough - I don't think it even needs to be 'orig_request'.

jupyter_server_proxy/handlers.py
```diff
@@ -255,22 +297,62 @@ async def proxy(self, host, port, proxied_path):
...
+            # Initial value for rewriting
+            rewritten_response = original_response
+
+            for rewrite in rewrite_responses:
+                # The rewrite function is a function of the RewritableResponse object
+                # ``response`` as well as several other optional arguments. We need to
+                # convert it to a function of only ``response`` by plugging in the
+                # known values for all the other parameters. (This is called partial
+                # evaluation.)
+                def rewrite_pe(rewritable_response: RewritableResponse):
```
yuvipanda, line 331: Can you use `functools.partial` here?

yuvipanda (approved):
Some style suggestions and a note about reducing the API surface further, but I am pretty happy with this! Thank you for working on this and incorporating my feedback, @maresb!!!!

yuvipanda:
> Sorry, I lost track of @yuvipanda's comment in the previous PR regarding `call_with_asked_args`. Now his request makes a lot more sense...

I was very confused until i read this haha :D Am glad you found it. Thank you for coming back and pushing this PR to completion.

jupyter_server_proxy/handlers.py
```diff
@@ -255,22 +297,62 @@ async def proxy(self, host, port, proxied_path):
...
+
+            # If this is a single function, wrap it in a list.
+            if isinstance(rewrite_response, (list, tuple)):
+                rewrite_responses = rewrite_response
+            else:
+                rewrite_responses = [rewrite_response]
+
+            # To be passed on-demand as args to the rewrite_response functions.
+            optional_args_to_rewrite_function = {
+                'orig_request': req,
```
ryanlovett, line 315: Hi @maresb, regarding @yuvipanda's [comment](https://github.com/jupyterhub/jupyter-server-proxy/pull/301#discussion_r746226931), could `req` here be changed to `self.request`? Knowing the request that the client made would help jupyter-rsession-proxy since it needs to know the client's requested URI as opposed to the one we have inside the proxy request `req`.

ryanlovett: @maresb As context, there's [an issue](https://github.com/jupyterhub/jupyter-rsession-proxy/issues/97) with RStudio where, in some jupyterhub environments, it redirects people from `/user/so-and-so/rstudio` to `/auth-sign-in` rather than `/user/so-and-so/rstudio/auth-sign-in`. Since `/auth-sign-in` hits jupyterhub and not the single user user, this is a 404. With your rewrite mechanism, jupyter-rsession-proxy can change the Location header, but it needs to know about `/user/so-and-so/rstudio` path, which I think is only in self.request. Let me know if you can think of a better way however.

ryanlovett: Oh, and per Yuvi's [other comment](https://github.com/jupyterhub/jupyter-server-proxy/pull/301/files#r749024114) maybe it could be `'request': self.request`.

yuvipanda (approved):
@ryanlovett does this look enough to support our rstudio workflow?

@maresb this looks great to me! A couple tests to test setting status and headers, and am ready to merge this. THANK YOU SO MUCH!

maresb:
Great! I didn't have as much time this weekend as I expected. I'll try to find some time within the next day or two to bring this over the finish line.

ryanlovett:
@yuvipanda Yes, I have a branch of jupyter-rstudio-proxy that works with @maresb latest on a local hub and on mybinder.

@maresb I'm also very thankful for this PR. Many downstream folks will be very happy!

minrk (approved):

ryanlovett:
Created https://github.com/maresb/jupyter-server-proxy/pull/1 which adds tests for status and headers to this PR.

ryanlovett (approved):

yuvipanda:
@maresb thank you for this wonderfully designed PR!

consideRatio:
I marked this as maintenance (reworks the function signature of the `rewrite_response` hook). Since the rewrite_response functionality hasn't been released, I don't mark this as breaking which it otherwise would have been I think.

PR #304, opened 2021-11-24
Set reason according to response code

maresb:
I noticed an annoying discrepancy between Tornado's `HTTPResponse` and my `RewritableResponse` when changing the HTTP code. In Tornado, the `.reason` attribute defaults to the text corresponding to the `.code` attribute. However, in `RewritableResponse`, we expect the `.code` attribute to be changed. If you change `.code`, then `.reason` retains the text generated for the previous code. In this PR, I add an observer which updates the `.reason` attribute if `.code` has changed and `.reason` still has its default value.

I also extended the documentation to illustrate more of the functionality provided, and extended the tests accordingly.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -56,11 +56,16 @@ class RewritableResponse(HasTraits):
 
     @observe('code')
     def _observe_code(self, change):
-        # Automatically update reason if code changes and previous reason was default.
-        old_default_reason = httputil.responses.get(change['old'], 'Unknown')
-        new_default_reason = httputil.responses.get(change['new'], 'Unknown')
-        if self.reason == old_default_reason:
-            self.reason = new_default_reason
+        # HTTP status codes are mapped to short descriptions in the
+        # httputil.responses dictionary, 200 maps to "OK", 403 maps to
+        # "Forbidden" etc.
+        #
+        # If code is updated and it previously had a reason matching its short
+        # description, we update reason to match the new code's short
+        # description.
+        #
+        if self.reason == httputil.responses.get(change['old'], 'Unknown'):
+            self.reason = httputil.responses.get(change['new'], 'Unknown')
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -17,7 +17,7 @@ from tornado import gen, web, httpclient, httputil, process, websocket, ioloop, 
 from jupyter_server.utils import ensure_async, url_path_join
 from jupyter_server.base.handlers import JupyterHandler, utcnow
 from traitlets.traitlets import HasTraits
-from traitlets import Bytes, Dict, Instance, Integer, Unicode, Union, default
+from traitlets import Bytes, Dict, Instance, Integer, Unicode, Union, default, observe
 
 from .utils import call_with_asked_args
 from .websocket import WebSocketHandlerMixin, pingable_ws_connect
@@ -53,6 +53,20 @@ class RewritableResponse(HasTraits):
     @default('reason')
     def _default_reason(self):
         return self.orig_response.reason
+
+    @observe('code')
+    def _observe_code(self, change):
+        # Automatically update reason if code changes and previous reason was default.
+        old_default_reason = httputil.responses.get(change['old'], 'Unknown')
+        new_default_reason = httputil.responses.get(change['new'], 'Unknown')
+        if self.reason == old_default_reason:
+            self.reason = new_default_reason
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Trigger the default value to be set from orig_response on instantiation.
+        # Otherwise _observe_code will receive change['old'] == 0.
+        self.code
 
     def _apply_to_copy(self, func):
         """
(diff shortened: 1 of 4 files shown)


jupyter_server_proxy/handlers.py
```diff
@@ -54,6 +54,20 @@ def _default_code(self):
     def _default_reason(self):
         return self.orig_response.reason
 
+    @observe('code')
+    def _observe_code(self, change):
+        # Automatically update reason if code changes and previous reason was default.
+        old_default_reason = httputil.responses.get(change['old'], 'Unknown')
+        new_default_reason = httputil.responses.get(change['new'], 'Unknown')
+        if self.reason == old_default_reason:
+            self.reason = new_default_reason
```
consideRatio, line 63: Is this really correct?

I think `change["old"]` reflects the old value of the `code` propery, and `change["new"]` reflects the new value of the code property. Based on the variable names, it seems like they are populated with the wrong content, and therefore the if statement below does something weird as well.

I think as part of a function observing `code`, you can only reason about before/after values of `code`, not before/after values of reason.

It seems to me like the variables are assigned a old / new value of the `code` propery, and that this is compared to the `reason` property. If you have logic comparing something with reason, shouldn't it be a comparison between reasons?

maresb: Thanks @consideRatio for taking a look!

> Based on the variable names, it seems like they are populated with the wrong content

Hmm, it looks correct to me, I don't understand what you mean. Could you perhaps spell it out more explicitly?

I'll write out an example here of how this logic is supposed to work. Hopefully that will eliminate confusion...

maresb: Let's say the initial response has `code=200` and `reason="OK"`. If I rewrite the response so that `code=404`, then I want `reason` to automatically switch to `404` as long as `reason` was not modified. In this example,

```python
change['old'] = 200
old_default_reason = "OK"
change['new'] = 404
new_default_reason = "Not Found"
```
Then the if statement checks that `self.reason == "OK"` and adjusts it to `self.reason = "Not Found"`.

The effect is that `reason` remains consistent if one only adjusts `code`. But one can still adjust `reason` independently.

Does this make sense? Is my code actually consistent with what I'm saying? (Maybe I'm blind to some typo I made.)

consideRatio: If you want to set reason to the response code, then you are doing it - practically. But, consider for example...

```
        old_default_reason = httputil.responses.get(change['old'], 'Unknown')
```

Isn't this really just setting a variable named `old_default_reason` to the old response code? Shouldn't you name the variable `old_response_code` then and let your logic act based on that?

I think you are making an implicit assumption that the response reason is the response code, but it seems weird for me that you would have it like that.

maresb: > It seems to me, that you don't want the reason to be the same as the code?

Well, `code` is an integer, and `reason` is a textual description which should correspond to that code. (And `httputil.responses` is a dictionary mapping numerical codes to textual descriptions.) Since `code`s are `int`s and `reason`s are `str`s, I don't understand how they could be the same.

I do want the reason to _correspond_ to the code, unless a custom string is being provided for `reason`.

Perhaps another example would help? I want to be able to do

```python
response.reason = "The server is sleepy"
response.code = 500
```
or equivalently
```python
response.code = 500
response.reason = "The server is sleepy"
```

I agree that things are confusing, although I don't yet understand your particular confusion. I think the jsp code would greatly benefit from type annotations. :smile:

consideRatio: > (And httputil.responses is a dictionary mapping numerical codes to textual descriptions.)

This was my confusion, I'm not sure what I thought... I now understand it very differently!

Is it correct you want reason to be based on the code, and kept updated based on the code, unless someone has explicitly set the reason?

consideRatio: Based on the comment "Automatically update reason if code changes and previous reason was default.", I think this is the code I'd write.

```python
    def _observe_code(self, change):
        # Automatically update reason if code changes and previous reason was default.
        if self.reason == self._default_reason():
            self.reason = httputil.responses.get(change['new'], 'Unknown')
```

But, I think you may be looking for logic to automatically update the reason if a user has explicitly set it in a way that isn't generated based on the code? Hmmm... I would try write this a bit differently. Maybe this would work for example, and then you remove the entire `def _observe_code(self, change):` function?

```python
    @default('reason')
    def _default_reason(self):
        return httputil.responses.get(self.code, 'Unknown')
```

consideRatio: What about...

```suggestion
        # HTTP status codes are mapped to short descriptions in the
        # httputil.responses dictionary, 200 maps to "OK", 403 maps to
        # "Forbidden" etc.
        #
        # If code is updated and it previously had a reason matching its short
        # description, we update reason to match the new code's short
        # description.
        #
        if self.reason == httputil.responses.get(change['old'], 'Unknown'):
            self.reason = httputil.responses.get(change['new'], 'Unknown')
```

maresb:
Thanks @consideRatio, your comment is very nicely written.

meeseeksmachine:
This pull request has been mentioned on **Jupyter Community Forum**. There might be relevant details there:

https://discourse.jupyter.org/t/a-template-repository-for-running-latest-rstudio-r-on-mybinder-org-your-jupyterhub/12235/1

PR #310, opened 2021-11-29
ci: avoid triggering ci twice on pre-commit.ci/dependabot prs

consideRatio:


.github/workflows/test.yaml
```diff
@@ -1,29 +1,30 @@
...
-  run:
-    # Declare bash be used by default in this workflow's "run" steps.
-    #
-    # NOTE: bash will by default run with:
-    #   --noprofile: Ignore ~/.profile etc.
-    #   --norc:      Ignore ~/.bashrc etc.
-    #   -e:          Exit directly on errors
-    #   -o pipefail: Don't mask errors from a command piped into another command
-    shell: bash
-
```
consideRatio, line 26: This is just something silly that wasn't needed, and has propagated across various github workflow files because i got it wrong initially. This can be relevant if we have a workflow job running in a windows machine and a linux machine - not something we have ever done.

PR #312, opened 2021-11-29
Remove no longer needed logic involving six

consideRatio:


jupyter_server_proxy/utils.py
```diff
@@ -44,7 +43,7 @@ class Callable(TraitType):
     info_text = 'a callable'
 
     def validate(self, obj, value):
-        if six.callable(value):
+        if callable(value):
```
consideRatio, line 46: This code is copy-pasted from https://github.com/ipython/traitlets/blob/a1425327460c4a3ae970aeaef17e0c22da4c53c6/traitlets/traitlets.py#L3232-L3246, where the old code included six, but the new code doesn't.

bollwyvl (approved):
hooray removing undeclared deps. LGTM,

jupyter_server_proxy/utils.py
minrk: Yup, and `six.callable` was only ever necessary to support Python 3.0 and 3.1 - it was in 2.7, removed in 3.0, and put back in 3.2.

PR #313, opened 2021-11-30
Modernize docs without making changes to its content

consideRatio:
- docs: update Makefile with devenv and linkcheck
- docs: update conf.py to not clutter as much
- docs: prepare for but don't switch to pydata theme yet
- docs: add linkcheck test
- docs: update links for the new linkcheck test
- Update .gitignore


docs/conf.py
```diff
@@ -1,32 +1,19 @@
...
-author = u'Yuvi Panda'
+# -- Project information -----------------------------------------------------
 
-# The short X.Y version
-version = u''
-# The full version, including alpha/beta/rc tags
-release = u'1.0'
+project = "Jupyter Server Proxy"
+copyright = f"{datetime.date.today().year}, Project Jupyter Contributors"
+author = "Project Jupyter Contributors"
```
consideRatio, line 16: @yuvipanda do you think this the right call of me to make, to update this like that?

You and Ryan are listed as authors of this package via a package.json that is read by setup.py as well still.

```json
  "author": {
    "name": "Ryan Lovett & Yuvi Panda",
    "email": "rylo@berkeley.edu"
  },
```

yuvipanda: Fine by me here. Should ping ryan too for the other bit :)

consideRatio: :+1: I'm not making that change!

yuvipanda:
Thanks, @consideRatio

PR #314, opened 2021-11-30
Remove empty JupyterLab style

### LINKED ISSUE #308: Remove (or add) JupyterLab style
### Proposed change

As of #307, the built `jupyter_server_proxy/labextension/static/third-party-licenses.json` lists `css-loader` and `style-loader`, but `jupyterlab-server-proxy/style/index.css` is empty, so these are not providing much value. So the question is: 

> Do we need any style, and/or would removing it in the future make things more complicated?

### Alternative options

- [ ] delete `index.css` and the reference to it in `index.ts`
- [ ] put something in it

If removed, this could be a quick yard stick of whether a PR is overcomplicating things, as the best run-time dependencies are _no_ run-time dependencies.

We could then try to _keep_ from shipping any third-party code until it's absolutely necessary. This can be figured out, roughly, by looking at the built `jupyter_server_proxy/labextension/static/third-party-licenses.json`.

Some additional style _could_ eventually be nice. If we ever supported log viewing, etc. it would possibly become necessary and better to write it in well-behaved `.css` (where it can be linted, tested, etc) than in fly-by-the-seat-of-your-pants hot style modifications on the frotnend as seems to be popular. I'll _always_ fight against styled-components or whatever that generate classes like `sc-asndicunas sc-easdasd`, and are not stable over time, not matter how nice they are for the developer experience or tiny performance gains.

bollwyvl:
- fixes #308
- removes empty `index.css` and reference to it in `index.ts`
  - by extension, removes the need to ship `style-loader` and `css-loader` runtimes
  - `third-party-licenses.json` will still be created, but will be empty

### DIFF
--- a/jupyterlab-server-proxy/src/index.ts
+++ b/jupyterlab-server-proxy/src/index.ts
@@ -3,8 +3,6 @@ import { ILauncher } from '@jupyterlab/launcher';
 import { PageConfig } from '@jupyterlab/coreutils';
 import { IFrame, MainAreaWidget, WidgetTracker } from '@jupyterlab/apputils';
 
-import '../style/index.css';
-
 function newServerProxyWidget(id: string, url: string, text: string): MainAreaWidget<IFrame> {
   const content = new IFrame({
     sandbox: ['allow-same-origin', 'allow-scripts', 'allow-popups', 'allow-forms'],


yuvipanda:
Thanks @bollwyvl!

Issue #322, opened 2022-02-15
jupyter-server-proxy lab extension cannot be disabled


cmd-ntrf:
### Bug description

After running the jupyter labextension disable command on jupyter-server-proxy, the extension is still being loaded.
```
jupyter labextension disable @jupyterlab/server-proxy
```

#### Expected behaviour

The server extension should still be loaded, but the frontend extension should not be activated when loading jupyterlab.

#### Actual behaviour

The jupyterlab extension is activated regardless of the enable/disable status of jupyter-server-proxy labextension status.

### How to reproduce

In a virtual environment:
1. `pip install jupyterlab jupyter-server-proxy jupyter-rsession-proxy`
2. `jupyter labextension disable @jupyterlab/server-proxy`
3. `jupyter lab`
4. The RStudio button still shows up in Jupyterlab launcher.

### More details

The reason why I want to disable the jupyterlab extension while maintaining the server extension is because I have an extension that handles the server proxy launcher items instead of jupyter-server-proxy (i.e: [jupyter-lmod](https://github.com/cmd-ntrf/jupyter-lmod)).

I suspect the issue stems from the way jupyter-server-proxy install its lab extension using setup.py instead of using the `jupyter labextension` commands. What leads me to think that is that if I try to uninstall just the lab extension of jupyter-server-proxy like this:
```
$ jupyter labextension uninstall @jupyterlab/server-proxy
```

I am greeted with the following message:
```
JupyterLab cannot uninstall @jupyterlab/server-proxy since it was installed outside of JupyterLab. Use the same method used to install this extension to uninstall this extension.
```

bollwyvl:
A key piece is what _major_ version of JupyterLab is involved.

`labextension uninstall` is for so-called "source extensions", which get built into the jupyterlab application itself with webpack and nodejs, and were the the only kind supported in JupyterLab 2.

In JupyterLab 3, which introduced "pre-built" extensions installable via `pip` or `conda`, one just needs to update some settings in the JupyterLab  configuration rather than rebuilding the whole application. For more see  the [`jupyter labextension disable`](https://jupyterlab.readthedocs.io/en/stable/user/extensions.html#enabling-and-disabling-extensions) command.

cmd-ntrf:
As per the provided reproduction steps, the problem was experienced with the latest version of jupyterlab - jupyterlab 3.2.9.

bollwyvl:
[...]

PR #323, opened 2022-02-16
Rename jupyterlab extension id to match convention

cmd-ntrf:
Allow disabling the extension using the name reported for the plugin by jupyter labextension CLI tools.

More details in issue #322

### DIFF
--- a/jupyterlab-server-proxy/src/index.ts
+++ b/jupyterlab-server-proxy/src/index.ts
@@ -104,7 +104,7 @@ async function activate(app: JupyterFrontEnd, launcher: ILauncher, restorer: ILa
  * Initialization data for the jupyterlab-server-proxy extension.
  */
 const extension: JupyterFrontEndPlugin<void> = {
-  id: 'jupyterlab-server-proxy',
+  id: '@jupyterlab/server-proxy:plugin',
   autoStart: true,
   requires: [ILauncher, ILayoutRestorer],
   activate: activate


consideRatio:
Important, this was considered a breaking change:

> The name should probably follow the upstream pattern, e.g. @jupyterlab/server-proxy:plugin, but shouldn't be changed until a major version bump of this package (e.g 4.0.0).

consideRatio:
Should this be named @jupyterlab/server-proxy to match the legacy npm package name though, or should it be named something more coupled to its current name - `jupyter-server-proxy` - the PyPI package?

yuvipanda:
I always imagine that extensions under the `@jupyterlab/` extension naming are in the jupyterlab org, rather than the jupyterhub org.

consideRatio:
This repo has a npm package in the jupyterlab org, but we dont need it for jupyterlab 3+ so should we reference it here now?

I think we want to reference a name including the repo name/pypi name and maybe also jupyterhub/ - but not the old npm package unless its really to point to some npm package - in whatever format is common.

bollwyvl:
This repo has a npm package in the jupyterlab org, but we dont need it for
> jupyterlab 3+ so should we reference it here now?
>
Yes, by convention it should keep publishing under an npm name, ideally the
same one. What org it is under isn't very important, but having it be
_something_ consistent is important for downstream packages.

consideRatio:
Hey, triaging open PRs and marking this as a draft.

I figure that unless we can name this @jupyterhub/jupyter-server-proxy that seems proper, we shouldn't make a breaking change about this.

If we can do that depends on getting control of the npm organization first and we are awaiting a claim to take control of an existing jupyterhub organization now in https://github.com/jupyterhub/team-compass/issues/562#issuecomment-1240714269.

consideRatio:
We have control of the jupyterhub organization, @minrk @manics what do you think - should we go for it?

I like the idea of decoupling this from being under @jupyterlab and named inconsistently like `server-proxy` instead of `jupyter-server-proxy` as the repo and python package is named.

Is it okay if I move towards resolving this by defining a NPM package / setting up credentials etc for the CI system to push to @jupyterhub/jupyter-server-proxy going onwards?

consideRatio:
Closing in favor of #363. We need to make more significant changes than just the one-liner. If there is agreement in this linked github issue, I'll go for it.

PR #339, opened 2022-06-21
Accept an unset `command` to proxy to an already started process (unmanaged process)

ryshoooo:
## Updated PR description by Erik

jupyter-server-proxy enables users to proxy to already started processes by `{base_url}/proxy/{port}/{proxied_path}`, but for a better user experience it could be nice to just be able to use `{base_url}/{proxied_path}`. This would rely on someone having configured how that should be proxied ahead of time. This PR adds support for this.

By providing a server process configuration like this, omitting the previousy required `command`, jupyter-server-proxy will assume the server process isn't to be managed/started by jupyter-server-proxy, and that its already started and accepting traffic. jupyter-server-proxy won't try to await it either before proxying traffic to it.

```python
# Below we configure jupyter-server-proxy to
# provide the route {base_url}/already-started-process
# as a convenience for the user un-friendly route
# {base_url}/proxy/54321/already-started-process,
# note how command is omitted!
c.ServerProxy.servers = {
{
    "already-started-process": {
        "port": 54321,
    },
}
```

Not providing a command to a server process definition means that its an "jupyter-server-proxy unmanaged server process" that we proxy to, supposed to be started ahead of time.

## Outdated PR description

This PR introduces additional parameter for the server proxy configuration, which specifies whether the port availability checking is necessary to perform or not. This is useful to disable if the application is already running (started by something else) and the user just wants to create a simple proxy for it (not start it through proxy directly).


ryanlovett:
Hi @ryshoooo, if/when I need to start the proxied service manually I use the anonymous proxy URL, e.g. `{base_url}/proxy/{port}/{proxied_path}` rather than `{base_url}/{name}/{proxied_path}`. Is this method not desired or sufficient? Is this what you mean by, "(not start it through proxy directly)" ?

ryshoooo:
> Hi @ryshoooo, if/when I need to start the proxied service manually I use the anonymous proxy URL, e.g. `{base_url}/proxy/{port}/{proxied_path}` rather than `{base_url}/{name}/{proxied_path}`. Is this method not desired or sufficient? Is this what you mean by, "(not start it through proxy directly)" ?

Hi @ryanlovett, yes it is exactly not desired :) To give you a specific example of what we're facing, we're using Z2JH with Kubernetes and we wish to provide code-server for each single-user environment. The old way we used to do this, was that we pre-installed code-server into the single-user base image and then used j-s-p the way it's supposed to be used, i.e. the command starts code-server and we open it up at some random open port. We're moving away from this solution in order to have even more micro-service architecture in place, so now we start the code-server as an additional extra container in the single-user pod during spawning. This approach has quite a few advantages, mainly that now we can do upgrades of code-server without touching the single-user base image. Anyway, the users got very much used to the fact that URL `user/{username}/code` points to the vscode, so we'd like to preserve that with the new setup :)

manics:
I think the general idea sounds good. Did you consider any other options, e.g. something like `managed_service=True|False` to control whether JSP starts the service or not, which is easier to debug (since you're not in an indeterminate state)

manics:
> But then it makes command parameter conditionally optional, which I'm not sure how I feel about it, feels like spaghetti.

That's a good point! Maybe we could make the command argument optional- if it's empty just setup the port-fowarding?
@ryanlovett what do you think?

ryanlovett:
@manics That seems reasonable to me.

ryshoooo:
I've added the changes, removed the `check_port` parameter, and made the command argument optional. However, the tests seem to be failing due to missing `jupyter-notebook` binary? Seems a bit unrelated to this PR right?

ryanlovett:
Thanks @ryshoooo ! Yes, I'm guessing we need to add `jupyter-notebook` as a new dependency before running tests.

manics:
The tests should be fixed by https://github.com/jupyterhub/jupyter-server-proxy/pull/340

Would you mind adding a test for this new configuration, e.g. based on
https://github.com/jupyterhub/jupyter-server-proxy/blob/099fce21feed4c5c762740dea2d34c5ca548c5b3/tests/resources/jupyter_server_config.py#L39-L41
https://github.com/jupyterhub/jupyter-server-proxy/blob/099fce21feed4c5c762740dea2d34c5ca548c5b3/tests/test_proxies.py#L141-L150

For the backend service you can just connect to one of the existing ones, so effectively the new configuration would be an alias.

ryshoooo:
@manics 

Thanks for the guidance :) I've added an additional test for the new functionality.

ryshoooo:
Looks like the pipeline is failing again, but the logs are a bit vague. Does anybody have any idea what's going on?

docs/source/server-process.rst
```diff
@@ -36,7 +36,8 @@ pairs.
    * A callable that takes any :ref:`callable arguments <server-process/callable-arguments>`,
      and returns a list of strings that are used & treated same as above.
   
-   This key is required.
+   If the command is not specified or is an empty list, it is assumed by the
+   proxy that the process is already running and available on ``{port}``.
```
consideRatio, line 40: I'm updating the language here with two goals:

1. "by the proxy" is a too vague reference to jupyter-server-proxy and better omitted
2. referencing that it is available on `port` could be nice, but being that explicit makes this PR conflict with #337 so I'm omitting it to make this documentation not need to be updated if that gets merged.

```suggestion
   If the command is not specified or is an empty list, the server process is
   assumed to be started ahead of time and already available to be proxied to.
```

jupyter_server_proxy/config.py
```diff
@@ -152,12 +153,16 @@ class ServerProxy(Configurable):
...
-            A list of strings that should be the full command to be executed.
+            An optional list of strings that should be the full command to be executed.
             The optional template arguments {{port}} and {{base_url}} will be substituted with the
             port the process should listen on and the base-url of the notebook.
 
             Could also be a callable. It should return a list.
 
+            If the command is not specified or is an empty list, it is assumed the process is already
+            running, therefore the port checking is skipped and the proxy is set up on the specified
+            port.
```
consideRatio, line 164: Reflects the change above, and avoids the technical language of "port checking" which also makes the assumption that we don't use a unix socket as could be introduced by #337.

```suggestion
            If the command is not specified or is an empty list, the server
            process is assumed to be started ahead of time and already available
            to be proxied to.
```

jupyter_server_proxy/handlers.py
```diff
@@ -650,6 +654,9 @@ async def ensure_process(self):
                 proc = SupervisedProcess(self.name, *cmd, env=server_env, ready_func=self._http_ready_func, ready_timeout=timeout, log=self.log)
                 self.state['proc'] = proc
 
+                if not cmd:
+                    return
+
```
consideRatio, line 659: Is this called to late? Should `self.state['proc']` be set to a new `SupervisedProcess` object? The idea was that if the command is unset, it shouldn't be supervised - right?

If it is correct, a comment is needed to clarify why it is correct in my mind as this is quite unexpected for me.

consideRatio: Besides this, the changes looks good to me!

consideRatio:
> Looks like the pipeline is failing again, but the logs are a bit vague. Does anybody have any idea what's going on?

Please ignore them. I'm not sure, but its related to jupyter_server 2 being released and that the jupyterlab extension isn't doing well with it.

jupyter_server_proxy/handlers.py
consideRatio: Like it is now, it won't be awaited readiness either. I'll push a commit to move this logic upwards.

consideRatio (approved):
This LGTM as it is now. I added two commits, one updating docs and another one relocating a return statement to happen sooner rather than later when an unmanaged process is involved.

- The test failures should be unrelated, see #362.

@ryshoooo does this look good to you as it is now?

ryshoooo:
> This LGTM as it is now. I added two commits, one updating docs and another one relocating a return statement to happen sooner rather than later when an unmanaged process is involved.
> 
> * The test failures should be unrelated, see #362.
> 
> @ryshoooo does this look good to you as it is now?

Hi @consideRatio 

Yes this looks totally fine to me, thanks a lot for your help! :)

consideRatio:
Thank you @ryshoooo! ❤️ 🎉

PR #340, opened 2022-07-08
Install `notebook<7` for notebook test

manics:
nbclassic 0.3.7 (required by jupyterlab) automatically installs notebook, but the recently released 0.4.2 doesn't. I think this is because nbclassic 0.4+ now includes all the frontend content that was previously in the notebook package. Since jupyter-server provides the backend for nbclassic it doesn't require notebook.


ryanlovett:
Thanks @manics!

PR #344, opened 2022-08-18
Fix config filename in config via traitlets doc

cmd-ntrf:
The file has to be named `jupyter_server_config.py` when using JupyterLab and version of Notebook >= 6.5. 

This is because jupyter-server is replacing the notebook backend:
https://jupyter-server.readthedocs.io/en/latest/operators/migrate-from-nbserver.html

It might be better to actually explain how the file should be named based on the version used, but I wanted to provide a starting point for the conversation.


hoogenm:
Useful info (correct filename for notebook >= 6.5).

consideRatio:
Thanks @cmd-ntrf!! :)

PR #348, opened 2022-11-01
ci: test against modern python versions and fix broken test suite

consideRatio:
I figure the most important versions to test are the oldest and newest. But,  it can help to know if something is just too old or just too new to function as well.

With this PR I update the CI system to run against py36, 37, 310, and 311, instead of py36, 37, 38, 39.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,7 +37,7 @@ jobs:
         # specified below, these parameters have no meaning on their own and
         # gain meaning on how job steps use them.
         jupyterlab_version: [2, 3]
-        python: [3.6, 3.7, 3.8, 3.9]
+        python: ["3.6", "3.11"]
         jupyter_app: [notebook, lab]
 
     steps:
--- a/tests/acceptance/Classic.robot
+++ b/tests/acceptance/Classic.robot
@@ -1,3 +1,7 @@
+*** Comments ***
+To learn more about these .robot files, see
+https://robotframework-jupyterlibrary.readthedocs.io/en/stable/.
+
 *** Settings ***
 Documentation     Server Proxies in Notebook Classic
 Library           JupyterLibrary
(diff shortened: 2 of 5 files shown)


bollwyvl:
`jupyter_core` might not pick up `jupyter_*_config.json` in their previous cwd locations... https://discourse.jupyter.org/t/security-advisory-for-jupyter-core-package/16527.

`--config-path` might work.

.github/workflows/test.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
         # specified below, these parameters have no meaning on their own and
         # gain meaning on how job steps use them.
         jupyterlab_version: [2, 3]
-        python: [3.6, 3.7, 3.8, 3.9]
+        python: ["3.6", "3.7", "3.10", "3.11"]
```
bollwyvl, line 40: 3.6 is EOL, probably just drop that, too. Might just need `3.7` and `3.11`.

consideRatio:
@bollwyvl I made a few attempts but can't understand how to pass `--config` flags to `jupyter server` via `__init__.robot`.

I attemted in 9f5f750 and a few other similar commits. I've read in [source code](https://github.com/robots-from-jupyter/robotframework-jupyterlibrary/blob/660ef6661aafd0c1f0f79e6a61443301831eb3df/src/JupyterLibrary/keywords/server.py#L16-L103) and [documentation](https://robotframework-jupyterlibrary.readthedocs.io/en/stable/KEYWORDS.html) without understanding what I can do here.

.github/workflows/test.yaml
consideRatio: I agree deal! I'd like to make a dedicated PR for dropping support for python 3.6 though.

consideRatio:
Thank you for helping out with this @bollwyvl!! I'm quite lost and has lots to learn about .robot stuff

consideRatio:
Wieeee! I failed to set `--config` but I managed to declare `JUPYTER_CONFIG_DIR` and indirectly do it. Thank you @bollwyvl for the correct guess on what was going wrong!!

consideRatio:
Merging and going for a support removal of py36 PR

PR #350, opened 2022-11-02
Drop support for Python 3.6

consideRatio:
For reference jupyterhub/jupyterhub has also dropped support for Python 3.6, and it has reached end of life 10 months ago. Requiring Python 3.7+ for the next release of this software seems very reasonable.


manics (approved):

consideRatio:
As this also was suggested by @bollwyvl in https://github.com/jupyterhub/jupyter-server-proxy/pull/348#discussion_r1011016380, I figure I'll go for a merge here and rebase #351 on it!

PR #352, opened 2022-11-03
docs: switch from alabaster to sphinx-book-theme, and misc docs build system maintenance

consideRatio:
I'm refreshing a few projects documentation build system.

In this case, it also meant leaving the old alabaster theme behind and going for sphinx-book-theme. A preview comparison is available.

- [Current docs](https://jupyter-server-proxy.readthedocs.org)
- [Docs PR preview](https://jupyter-server-proxy--352.org.readthedocs.build/en/352/)

## Related
- https://github.com/jupyterhub/zero-to-jupyterhub-k8s/pull/2909
- https://github.com/jupyterhub/binderhub/pull/1565
- https://github.com/jupyterhub/repo2docker/pull/1197


choldgraf (approved):
Definitely an improvement to me!

consideRatio:
Thanks @choldgraf! Going for a self-merge with that approval to enable me to rebase #353 on it and resolve a merge conflict that otherwise occur.

choldgraf:
This makes me wonder - does the jupyterhub team have a policy about self merging when there is an approving review? My feeling is that for most PRs, after one approving review we should encourage a merge by anybody unless there are "request changes" present but somebody else, so that you don't need to deal with the decision cost of knowing whether you can self merge

consideRatio:
Decision cost was a nice concept to relate to!

I agree on that policy I think. Maybe also along with a unburdening clarification for reviewers that that they could do partial reviews ending up in a request for help to understand something or only provide a partial approval that includes a disclaimer that they didn't understand the change details or consequences well enough and ask that another person contribute a review as well.

Pinging a related topic: https://github.com/jupyterhub/team-compass/issues/466

choldgraf:
That's what I use the "comment" review feature for. I think it's a helpful way to provide feedback while explicitly signaling that you don't wish to formally approve

PR #353, opened 2022-11-04
Refine RELEASE.md with tbump

consideRatio:
We have adopted `tbump` to do commit/tag/push in several other jupyterub org repos and I'm very positive about the experience so far even though I was quite skeptical initially.


consideRatio:
Merging to be able to reference the new RELEASE.md from another PR, hoping for help regarding the instructions on updating the yarn.lock.

Issue #358, opened 2022-11-19
Server Proxy does not forbid users to access web apps of other users running on same host


mahendrapaipuri:
Hello,

It seems like Jupyter server proxy does not stop one user to proxy to web application of another user running on the same host. If the web app does not have any authentication, any user can proxy to app of any other using just by knowing on which port the web app is running.

**Reproducer:**

We can reproduce it in container setup. Build a container with following `Dockerfile`

```
FROM debian:stable-slim

ARG DEBIAN_FRONTEND=noninteractive

# Install prerequisites
RUN apt-get update -y && apt-get install apt-utils -y && \
                         apt-get install vim \
                                         git \
                                         curl \
                                         python3.9 \
                                         python3-pip -y
                                         
# Install node
RUN curl -fsSL https://deb.nodesource.com/setup_18.x | bash - && \
    apt-get install -y nodejs

# Add users
RUN useradd --shell /bin/bash --home-dir /home/user1 -p $(openssl passwd -1 user1) --create-home user1 && \
    useradd --shell /bin/bash --home-dir /home/user2 -p $(openssl passwd -1 user2) --create-home user2

# Install Jupyterlab and jupyter server proxy
RUN pip install --no-cache-dir jupyterlab==3.5.0 jupyter_server_proxy==3.2.2 tensorboard==2.11.0

# Expose port
EXPOSE 8888

# Entrypoint
CMD ["bash"]
```

Now start a  `jupyterlab` instance as user `user1`

```
docker run -p 9999:8888 --rm -it --name jsp_test jsp
root@b20fcea9b498:/# su - user1
user1@b20fcea9b498:~$ jupyter-lab --ip=0.0.0.0
```

Now open another shell as start a `tensorboard` instance as `user2`

```
docker exec -it jsp_test bash
root@b20fcea9b498:/# su - user2
user2@b20fcea9b498:~$ tensorboard --logdir /home/user2 --port 7001
```

Now if we go to `http://localhost:9999/proxy/7001/`, we can access the `tensorboard` instance of `user2`

![Screenshot 2022-11-19 at 19 49 43](https://user-images.githubusercontent.com/44365948/202866792-d05f83fd-c336-4f7c-8bc5-f29ad2b7a089.png)

**Use case:**

This can be a potential security issue on multi tenant systems like HPC platforms. This is a very common scenario in HPC environment that multiple users run their `jupyterlab` instances on the same host.

**Potential Fix:**

[...]

PR #359, opened 2022-11-19
README: warn about unlimited local network access

manics:
I thought there was already a warning in this repo, but it doesn't look like it. For comparison:
https://github.com/jupyterhub/jupyter-rsession-proxy#multiuser-considerations has a warning.

Closes https://github.com/jupyterhub/jupyter-server-proxy/issues/358

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,8 +32,25 @@ The primary use cases are:
 [The documentation](https://jupyter-server-proxy.readthedocs.io/)
 contains information on installation & usage.
 
-**Warning:** Jupyter Server Proxy is designed for containerised environments.
-It has access to all local network services that the user has access to.
+## Security warning
+
+Jupyter Server Proxy is often used to start a user defined process listening to
+some network port (http://localhost:4567) for a user starting a Jupyter Server
+that only that user has permission to access. The user can then access the
+started process proxied through the Jupyter Server.
+
+For safe use of Jupyter Server Proxy, you should ensure that the process started
+by Jupyter Server proxy can't be accessed directly by another user and bypass
+the Jupyter Server's authorization!
+
+A common strategy to enforce access proxied via Jupyter Server is to start
+Jupyter Server within a container and only allow network access to the Jupyter
+Server through to the container. Another strategy is to communicate with the
+started process through a unix socket as proposed in [Pull request
+#337](https://github.com/jupyterhub/jupyter-server-proxy/pull/337).
+
+For more insights, see [Ryan Lovetts comment about
+it](https://github.com/jupyterhub/jupyter-server-proxy/pull/359#issuecomment-1350118197).
 
 ## Install
 
--- a/README.md
+++ b/README.md
@@ -32,6 +32,9 @@ The primary use cases are:
 [The documentation](https://jupyter-server-proxy.readthedocs.io/)
 contains information on installation & usage.
 
+**Warning:** Jupyter Server Proxy is designed for containerised environments.
+It has access to all local network services that the user has access to.
+
 ## Install
 
 ### Requirements
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ contains information on installation & usage.
 ## Security warning
 
 Jupyter Server Proxy is often used to start a user defined process listening to
-some network port (http://localhost:4567) for a user starting a Jupyter Server
+some network port (e.g. http://localhost:4567) for a user starting a Jupyter Server
 that only that user has permission to access. The user can then access the
 started process proxied through the Jupyter Server.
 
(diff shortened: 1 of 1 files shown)


consideRatio (commented):
We can merge this, but I think the warning clarification was tricky for me to grasp properly - this part:

> It has access to all local network services that the user has access to.

The description below was clear enough for me to feel that I understood it properly though.

>### Multiuser Considerations
>This extension launches an rstudio server process from the jupyter notebook server. This is fine in JupyterHub deployments where user servers are containerized since other users cannot connect to the rstudio server port. In non-containerized JupyterHub deployments, for example on multiuser systems running LocalSpawner or BatchSpawner, this not secure. Any user may connect to rstudio server and run arbitrary code.

In my own words, what would I say? Hmmm...

If jupyter-server-proxy is configured to start some application for a user and expose it on some port, its exposed for whoever can access that port - which typically are all users with network access to the computer or container. Or hmmm?

Hmmm, no I don't think I understand this well enough atm to describe the warning myself. Could you elaborate or make the warning more verbose @manics?

ryanlovett:
@consideRatio To elaborate a bit, jupyter-server-proxy starts applications which then usually listen on localhost ports. jupyter protects the path to `{jupyter_url}/application/` or `{jupyter_url}/port/{port}` so if anyone tries to access those URLs, they will not be able to auth. However the proxied application itself may not provide any additional authentication if a user connects to `localhost:{port}`. In a containerized environment this doesn't matter very much because only jupyter is exposed by the container. On a multiuser system however, if someone starts an application (through jupyter or otherwise) that is listening on `localhost:{port}`, another user on that system can connect to that same port without intervention. For example if user1 launches `cd ~/secrets/; python -m http.server 9000`, user2 can `curl localhost:9000` and read user1's secrets.

Users may launch jupyter on multiuser systems from the command-line, or through JupyterHub when it uses spawners like batchspawner or systemdspawner.

Potential fixes for this:
- https://github.com/jupyterhub/jupyter-server-proxy/pull/337, which exposes the proxied app through a UNIX socket instead of an open port
- https://github.com/ryanlovett/portwrap, which runs the proxied app in a user/network namespace so that jupyter can connect to the open port and proxy it, but nothing outside of the namespaces can reach that port

consideRatio:
Ah its a lot clearer to me now @ryanlovett, thanks!

consideRatio (commented):
With the code suggestion I made, this LGTM! Feel free to adjust it, but I think its relevant to clarify it to this degree or similar as part of mentioning it.

README.md
```diff
@@ -32,6 +32,26 @@ The primary use cases are:
 [The documentation](https://jupyter-server-proxy.readthedocs.io/)
 contains information on installation & usage.
 
+## Security warning
+
+Jupyter Server Proxy is often used to start a user defined process listening to
+some network port (http://localhost:4567) for a user starting a Jupyter Server
```
manics, line 38:
```suggestion
some network port (e.g. http://localhost:4567) for a user starting a Jupyter Server
```

README.md
```diff
@@ -32,6 +32,26 @@ The primary use cases are:
...
+
+For safe use of Jupyter Server Proxy, you should ensure that the process started
+by Jupyter Server proxy can't be accessed directly by another user and bypass
+the Jupyter Server's authorization!
+
+A common strategy to enforce access proxied via Jupyter Server is to start
+Jupyter Server within a container and only allow network access to the Jupyter
+Server through to the container. Another strategy is to communicate with the
+started process through a unix socket as proposed in [Pull request
+#337](https://github.com/jupyterhub/jupyter-server-proxy/pull/337).
```
manics, line 50:
```suggestion
Server via the container.
```

manics:
I took out the mention of the unix socket from the README since it's not merged, so it's not very helpful to a user! We can add it to the readme when it's merged.

consideRatio (approved):

im-not-an-alien-i-swear-whos-ur-leader:
@ryanlovett I like your suggestions, however, if jupyter is running in a container, what is stopping a user from using a command line in the jupyterlab instance and just accessing the hidden services in the container that way? It seems like you again have issues where you need service-specific auth such as a token, or unix sockets like you suggested.

I see that unix sockets are part of jupyter-server-proxy now so this is kind of a moot point, but I'm curious if I'm misunderstanding how the container would have helped.

Thanks

ryanlovett:
@im-not-an-alien-i-swear-whos-ur-leader Usually jupyter-based containers are single user only, for example launched by a JupyterHub. In this case, only the one user can access the listening services in their container.

Having multiple users launch different jupyter servers within the same container that also run proxied services is not great from a security perspective, but people traditionally use containers to isolate processes.

Issue #362, opened 2022-12-29
Tests broken for jupyter_server 2 - help wanted!


consideRatio:
Our jupyterlab extension tests for JupyterLab 2 and JupyterLab 3 are both failing when jupyter_server 2 is used, which it is in our Python 3.11 tests but not in our Python 3.7 tests because jupyter_server 2 requires python 3.8 and is not installed in the python 3.7 context.

# Summary to help understand the error

I've debugged this in a fork where I run tests in a Python 3.8 environment, and only switch between installing jupyter_server 1 and 2. Doing so, I conclude the following two separate errors we encounter when testing with jupyterlab 2 and 3 together with jupyter_server 2.

## JupyterLab 2 failure

When using jupyterlab 2 and jupyter_server 2 in a Python 3.8 environment, we fail in the `python -m jupyterlab.browser_check` step, and here are some diff from logs and python dependencies installed.

```diff
-# lab 2, py3.8, jupyter_server 1, notebook
+# lab 2, py3.8, jupyter_server 2, notebook
    app dir: /opt/hostedtoolcache/Python/3.8.15/x64/share/jupyter/lab
         @jupyterhub/jupyter-server-proxy v3.2.2  enabled  OK*
 JupyterLab v2.3.2
@@ -83,11 +83,21 @@ Starting Chrome Headless
 Navigating to page: file:///home/runner/.local/share/jupyter/runtime/nbserver-x-open.html
 Waiting for page to load...
 Waiting for application to start...
+/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/jupyter_server/base/handlers.py:240: UserWarning: The Tornado web application does not have an 'identity_provider' defined in its settings. In future releases of jupyter_server, this will be a required key for all subclasses of `JupyterHandler`. For an example, see the jupyter_server source code for how to add an identity provider to the tornado settings: https://github.com/jupyter-server/jupyter_server/blob/aa8fd8b3faf37466eeb99689d5555314c5bf6640/jupyter_server/serverapp.py#L253
+  warnings.warn(
+[E x BrowserApp] Error unpacking user from cookie: Extra data: line 1 column 2 (char 1)
+Error unpacking user from cookie: Extra data: line 1 column 2 (char 1)
+[W x BrowserApp] Clearing invalid/expired login cookie username-127-0-0-1-8889
[...]

PR #364, opened 2022-12-29
ci: use ubuntu-22.04, firefox and geckodriver now available

consideRatio:
As seen [here](https://github.com/actions/runner-images/blame/e9b59a540ed5e360dbdbc018830daa828f9ec067/images/linux/Ubuntu2204-Readme.md?plain=1#L164-L180), firefox and geckodriver as needed by browser simulation tests in our CI system is now available in the ubuntu-22.04 environment. So, due to that we upgrade to use it.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,11 +27,12 @@ on:
 
 jobs:
   test:
-    # FIXME: We need ubutnu-20.04 until firefox and geckowebdriver is available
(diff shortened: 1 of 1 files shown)


manics (approved):

manics:
Can we pin `jupyter-server<2` for the tests?

consideRatio:
@manics, I'll get that done if I can't resolve #362 soon!

PR #371, opened 2023-01-07
Adopt hatchling, plugins for python package build

### LINKED ISSUE #370: Modernize packaging infrastructure
### References

- follow-on to #368

### Proposed change

Update the packaging infrastructure to be in line with current common practices e.g. [cookiecutter](https://github.com/jupyterlab/extension-cookiecutter-ts). 

> I'd say "best practices," but i'm still not 100% convinced of the relatively extreme extensibility of `hatch` and the number of things it does _magically_... i actually _prefer_ the relatively bare-bones, but extremely stable, `flit`, and generally use that on my own personal and `$DAY_JOB` projects, with an external tool such as `make` or `doit`.

- maintainers and contributors _should_ be able to `pip install -e .[atest,docs]` and get a full working dev environment
  - (modulo perhaps `nodejs`, `geckodriver`)

- use the `hatch` build backend and plugins to normalize on [`pyproject.toml`](https://github.com/jupyterlab/extension-cookiecutter-ts/blob/3.0/%7B%7Bcookiecutter.python_name%7D%7D/pyproject.toml)
- hoist the `package.json` to the top level of the repo to simplify some of the path stuff
- only ship a single copy of the built assets in the `.whl`

### Alternative options

- do nothing: keep using `setuptools`, etc. with legacy `setup.py` build?
  - will keep working for a while, yet
- keep using `setup.py`, etc. ?
  - add a `pyproject.toml` that declares the build system?
  - move most of the metadata to `setup.cfg`

bollwyvl:
## References

- fixes #370

## Elevator Pitch

- [x] adopt `hatchling` and some plugins:
    - `hatch-jupyter-builder` handles the `jlpm` junk
    - `hatch-nodejs-version` fetches metadata from `package.json`
- [x] normalize dependencies and metadata
  - > `docs/requirements.txt` left to avoid node on RTD 
- [x] add CI caching
- [x] quick pass over user/dev docs
- [x] normalize the `etc` folders with their destination structure
  - > didn't do this for the nb/labextension due to path lengths (e.g. windows)


pyproject.toml
```diff
@@ -1,10 +1,114 @@
...
+[tool.hatch.version]
+source = "nodejs"
+path = "labextension/package.json"
+
+[tool.hatch.build.targets.sdist]
+artifacts = [
+    "jupyter_server_proxy/labextension",
+]
+exclude = [
+    ".git",
```
ofek, line 76: hey! just FYI this is always excluded

pyproject.toml
```diff
@@ -1,10 +1,114 @@
...
+    "robotframework-jupyterlibrary >=0.4.2",
+]
+docs = [
+    "myst-parser",
+    "sphinx-autobuild",
+    "sphinx-book-theme",
+    "sphinx-copybutton",
+    "sphinxext-opengraph",
+    "sphinxext-rediraffe",
+]
```
consideRatio, line 65: I did a similar change to introduce `docs` as a optional-dependencies target in another project in the jupyterhub org, and we agreed to just stick with the requirements.txt under docs/ so I figure we do this also in this project.

If RTD needs nodejs btw, it can be installed via configuration for us in the v2 version of their .readthedocs.yaml config file.

pyproject.toml
consideRatio, line 76:
```suggestion
```

pyproject.toml
consideRatio:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -108,25 +126,23 @@ jobs:
...
 
-      - name: Install Acceptance test dependencies
-        run: |
-          # the acceptance test requires notebook to run
-          pip install "notebook<7" "robotframework-jupyterlibrary>=0.4.2"
-
       - name: Run acceptance tests
         run: |
-          pytest -k acceptance -s
+          set -eux
```
consideRatio, line 138: I don't want to introduce something that isn't needed, and I don't think this should be needed.

1. -e --- We will already error by default in github actions shell scripts
2. -u --- About unset environment variables, I think maybe we won't error if we have an unset environment variable, but we don't provide on in this script section either.
3. -x --- GitHub actions already print stuff we do

If we do want to change something, we can do it in the `defaults.run` section I think.

consideRatio:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -108,25 +126,23 @@ jobs:
         # jupyterlab.browser_check with jupyterlab 2 and a modern version of
         # python (3.11+).
         #
-        if: ${{ !(matrix.jupyterlab-version == '2' && startsWith(matrix.python-version, '3.1')) }}
+        if: ${{ !(matrix.jupyterlab-version == '2' && startsWith(matrix.python-version, '3.11')) }}
         run: |
+          set -eux
```
consideRatio, line 131:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -43,61 +43,79 @@ jobs:
...
           pip install jupyter-server
           jupyter server extension list
-          jupyter server extension list 2>&1 | grep -ie "jupyter_server_proxy.*enabled" -
+          jupyter server extension list 2>&1 | grep -iE "jupyter_server_proxy.*OK" -
 
       - name: Install JupyterLab Extension
         if: matrix.jupyterlab-version == '2'
         run: |
+          set -eux
```
consideRatio, line 117:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -43,61 +43,79 @@ jobs:
...
-          name: unit-tests-${{ matrix.python-version }}-${{ matrix.jupyter-app }}-${{ matrix.jupyterlab-version }}-${{ github.run_number }}
+          name: |-
+            unit-tests-${{ matrix.python-version }}-${{ matrix.jupyter-app }}-${{ matrix.jupyterlab-version }}-${{ github.run_number }}
           path: |
             ./build/pytest
             ./build/coverage
 
       - name: Check the Notebook Server extension is installed
         run: |
+          set -eux
```
consideRatio, line 103:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -43,61 +43,79 @@ jobs:
...
+
+      - name: List Python packages
+        run: |
+          set -eux
           pip freeze
+          pip check
 
       - name: Run tests
         run: |
+          set -eux
```
consideRatio, line 85:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -43,61 +43,79 @@ jobs:
...
-          # Ensure we don't accidentally depend on notebook
-          if [ "${{ matrix.jupyter-app }}" == "notebook" ]; then
-            pip install "notebook<7"
-          fi
+          set -eux
+          pip install -vv $(ls ./dist/jupyter_server_proxy-*.whl)\[acceptance\] 'jupyterlab~=${{ matrix.jupyterlab-version }}.0'
+
+      - name: List Python packages
+        run: |
+          set -eux
```
consideRatio, line 79:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -43,61 +43,79 @@ jobs:
...
         run: |
-          pip install --upgrade pip
-          pip install jupyter_packaging jupyterlab~=${{ matrix.jupyterlab-version }}.0
-          pip install ./dist/jupyter_server_proxy-*.whl
-          pip install pytest pytest-cov pytest-html
-          # Ensure we don't accidentally depend on notebook
-          if [ "${{ matrix.jupyter-app }}" == "notebook" ]; then
-            pip install "notebook<7"
-          fi
+          set -eux
```
consideRatio, line 74:
```suggestion
```

.github/workflows/linkcheck.yaml
```diff
@@ -25,9 +25,11 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
+          cache: pip
+          cache-dependency-path: pyproject.toml
 
       - name: Install deps
-        run: pip install -r docs/requirements.txt
+        run: pip install .[docs]
```
consideRatio, line 32:
```suggestion
        run: pip install -r docs/requirements.txt
```

.github/workflows/linkcheck.yaml
```diff
@@ -25,9 +25,11 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
+          cache: pip
+          cache-dependency-path: pyproject.toml
```
consideRatio, line 29: I'm not so keen on caching because it often complicates debugging, even when it wasn't causing an issue as I may end up needing to rule out caching was an issue.

If we reference a lock file, it makes more sense to me though, but here we reference versions that are overall not very tightly pinned, so we want to test new versions as they arrive and fail just at the time when they arrive - and not when the cache times out at unknown time.

Example: I see that a failure is showing up in builds after 20th some month, but the new version of a package i wonder introduced a regression was released 16th. Without cache, I could say "ah, it must not have been that package released 16th because the fails showed up 20th and not 19th"

Due to that, I'd like us to remove the cache when we install from something else than a .lock file pinning dependencies super tight.

consideRatio, line 29:
```suggestion
```

.github/workflows/linkcheck.yaml
consideRatio: I comment about this below

.github/workflows/publish.yaml
```diff
@@ -35,14 +35,29 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+
       - uses: actions/setup-python@v4
         with:
           python-version: "3.11"
+          cache: pip
+          cache-dependency-path: pyproject.toml
```
consideRatio, line 43:
```suggestion
```

.github/workflows/publish.yaml
```diff
@@ -35,14 +35,29 @@ jobs:
...
+
+      - name: Update root build packages
+        run: |
+          pip install --upgrade build pip jupyterlab~=3.0
 
       - name: Build dist
         run: |
-          pip install jupyter_packaging wheel jupyterlab~=3.0
-          python setup.py sdist bdist_wheel
+          set -eux
```
consideRatio, line 58:
```suggestion
```

.github/workflows/publish.yaml
```diff
@@ -52,9 +67,12 @@ jobs:
 
       - name: Javascript package
         run: |
+          set -eux
```
consideRatio, line 70:
```suggestion
```

pyproject.toml
```diff
@@ -1,10 +1,114 @@
...
+    "simpervisor >=0.4",
+]
+
+[project.optional-dependencies]
+test = [
+    "pytest",
+    "pytest-cov",
+    "pytest-html",
+]
+acceptance = [
```
consideRatio, line 53:
```suggestion
# acceptance tests additionally require firefox and geckodriver
acceptance = [
```

CONTRIBUTING.md
```diff
@@ -39,15 +39,15 @@ To install these in addition to the [Python package](#python-package) test
 dependencies, run:
 
 ```bash
-pip install -e .[test,acceptance]
+pip install -e .[acceptance]
```
consideRatio, line 42: Depending on the shell, you may need quotes here I think.

```suggestion
pip install -e ".[acceptance]"
```

consideRatio (commented):
Thank you soo much for your work on this @bollwyvl!!!

Overall this LGTM but there was some changes I wished for with associated code suggestions.

1. Removal of `set -eux`
2. Removal of cache for python packages (but not for installing yarn.lock)
3. Revert addition of `jupyter-server-proxy[docs]` as a pip install target

bollwyvl:
In doint this, I'd recommend as well that we add a `.binder/{environment.yml,postBuild}` that exercises as much of this as possible, as well as the `contrib` stuff, but that can be another day.

.github/workflows/publish.yaml
```diff
@@ -35,14 +35,26 @@ jobs:
...
+      - uses: actions/setup-node@v3
+        with:
+          cache: yarn
+          node-version: 18.x
+          registry-url: https://registry.npmjs.org
+          cache-dependency-path: labextension/yarn.lock
+
+      - name: Update root build packages
+        run: |
+          pip install --upgrade build pip jupyterlab~=3.0
```
consideRatio, line 52: Is `jupyterlab~=3.0` needed to be specified here? I'm thinking that pyproject-build ought to handle this if its outlined as a requirement for the build env in pyproject.toml?

I tried and got this:

```
* Installing packages in isolated environment... (hatch-jupyter-builder >=0.5, hatch-nodejs-version, hatchling >=1.4.0, jupyterlab >=3.4.7,<4.0.0)
```

```suggestion
          pip install --upgrade build pip
```

consideRatio: Further down i see that we only have the following, so it seems correct to remove this part.

```
      - name: Update root build packages
        run: |
          pip install --upgrade build pip
      - name: Build Python package
        run: |
          pyproject-build
```

consideRatio (approved):
Besides the final comment, this LGTM!

Do you consider this ready for merge as well at this point @bollwyvl?

bollwyvl:
Yes, I've no further changes planned once it passes CI.

```
Run mkdir jsdist
  mkdir jsdist
  cd labextension
  jlpm pack --filename ../jsdist/labextension-jlpmpack.tgz
  cd ../jsdist && sha256sum * | tee SHA256SUMS
  shell: /usr/bin/bash -e {0}
/home/runner/work/_temp/395963e9-5a5b-4889-943b-08c64b11e3e6.sh: line 3: jlpm: command not found
```

I'll change this to just use `npm pack`.

.github/workflows/publish.yaml
bollwyvl: using `npm pack` (likely better, anyway, as then the as-built and as-uploaded SHA256 will always match... `yarn` does... different things)

bollwyvl:
Publish is passing, so this is probably good. Happy to make any other fixes, or let others do so... feel free to squash merge.

consideRatio:
Wiee nice thank you @bollwyvl!!! Super happy about this being modernized!!

I went for a normal merge as I think a few jupyterhub team members prefer we use that consistently.

PR #375, opened 2023-01-13
Fix imshields link README.md

prabhakk-mw:
Fixing imshields link based on https://github.com/badges/shields/issues/8671


manics:
Thanks for noticing this.

Are there any advantages in keeping the shields.io badge, instead of the ones provided by GitHub: 
`[Test](https://github.com/jupyterhub/jupyter-server-proxy/actions/workflows/test.yaml)`

prabhakk-mw:
No discernible difference, I guess the only advantage would be consistency as all the other badges are from `shields.io`

consideRatio:
@manics let's stay using shields.io for consistency, we can use it to provide links for everything and then also style them systematically etc.

PR #387, opened 2023-04-07
Drop support for Python 3.7

consideRatio:
Next release is 4.0.0, I figure we should drop support for Python 3.7 as part of it.

### Review input

Is there agreement on dropping Python 3.7 at this point before the 4.0.0 release?

### Test failures -> draft

Test failures remain, likeley caused by bumping test version from py37 to py38 allowed some newer version of software be installed where we previously could rely on only some old version was installed.


manics (approved):
We may need to modify the dependencies instllation further when notebook 7 (based on JupyterLab) comes out.

PR #392, opened 2023-04-10
Test Windows

manics:
Test https://github.com/jupyterhub/simpervisor/pull/26

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -39,11 +39,6 @@ jobs:
             jupyterlab-version: "3"
             os: windows-2022
 
-    # NOTE: The worker environment running this job needs firefox and
-    #       geckowebdriver available. Before upgrading to ubuntu-24.04 in the
-    #       future, check and see its available in a location like this:
-    #       https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers
-    #
     runs-on: ${{ matrix.os }}
     defaults:
       run:
@@ -86,7 +81,7 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -k "not acceptance" -vs
+          pytest -k "not acceptance" -vv
 
       - name: Upload pytest and coverage reports
         if: always()
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,15 +29,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        #         python-version: ["3.8", "3.11"]
-        #         jupyter_server-version: ["1", "2"]
-        #         jupyterlab-version: ["3"]
-        #         os: [ubuntu-22.04]
+        python-version: ["3.8", "3.11"]
+        jupyter_server-version: ["1", "2"]
+        jupyterlab-version: ["3"]
+        os: [ubuntu-22.04]
         include:
-          - python-version: "3.10"
-            jupyter_server-version: "2"
-            jupyterlab-version: "3"
-            os: windows-2022
           - python-version: "3.11"
             jupyter_server-version: "2"
             jupyterlab-version: "3"
@@ -90,7 +86,7 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -k "not acceptance" -vsx
+          pytest -k "not acceptance" -vs
 
       - name: Upload pytest and coverage reports
         if: always()
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,6 +25,7 @@ on:
 
 jobs:
   test:
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,10 +28,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.11"]
-        jupyter_server-version: ["1", "2"]
-        jupyterlab-version: ["3"]
-        os: [ubuntu-22.04]
+        #         python-version: ["3.8", "3.11"]
+        #         jupyter_server-version: ["1", "2"]
+        #         jupyterlab-version: ["3"]
+        #         os: [ubuntu-22.04]
         include:
           - python-version: "3.10"
             jupyter_server-version: "2"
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -47,7 +48,8 @@ dependencies = [
     "aiohttp",
     "importlib_metadata >=4.8.3 ; python_version<\"3.10\"",
     "jupyter-server >=1.0",
-    "simpervisor >=0.4",
+    # simpervisor >=1.0 for Windows support
+    "simpervisor",
```
manics, line 52:
```suggestion
    "simpervisor >=0.4",
```

consideRatio (approved):
Wieeeee nice @manics!!!

bollwyvl:
Perhaps the windows simpervisor version sounds like it should be more
formally pinned, e.g. with a `>=1.0 ; sys_platform` check in project,
especially if it's theoretically possible to install the older one there
and not get something working.

pyproject.toml
```diff
@@ -47,6 +48,7 @@ dependencies = [
     "aiohttp",
     "importlib_metadata >=4.8.3 ; python_version<\"3.10\"",
     "jupyter-server >=1.0",
+    # simpervisor >=1.0 for Windows support
     "simpervisor >=0.4",
```
manics, line 52:
```suggestion
    "simpervisor >=0.4 ; platform_system!=\"Windows\"",
    "simpervisor >=1.0 ; platform_system==\"Windows\"",
```

.github/workflows/test.yaml
```diff
@@ -76,7 +86,7 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -k "not acceptance"
+          pytest -k "not acceptance" -vs
```
bollwyvl, line 89: i'm generally pro verbosity (and recommend `-vv` for CI for better observability, especially if/when network things hang), but `-s` is often a bridge too far...

.github/workflows/test.yaml
```diff
@@ -25,19 +25,29 @@ on:
...
 jobs:
   test:
-    # NOTE: The worker environment running this job needs firefox and
-    #       geckowebdriver available. Before upgrading to ubuntu-24.04 in the
-    #       future, check and see its available in a location like this:
-    #       https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers
-    #
-    runs-on: ubuntu-22.04
-
+    timeout-minutes: 30
```
bollwyvl, line 28: this seems... like a lot... is it still needed? could individual offending tests be constrained with e.g. `pytest-timeout`?

.github/workflows/test.yaml
```diff
@@ -25,19 +25,29 @@ on:
...
         jupyter_server-version: ["1", "2"]
         jupyterlab-version: ["3"]
+        os: [ubuntu-22.04]
+        include:
+          - python-version: "3.11"
+            jupyter_server-version: "2"
+            jupyterlab-version: "3"
+            os: windows-2022
+
+    # NOTE: The worker environment running this job needs firefox and
```
bollwyvl, line 42: is this commend even accurate now?

bollwyvl:
This looks like really good stuff. What's the timeline for a real release of the relevant `simpervisor`?

Not strictly important, but on `conda-forge` to support the tighter dependency on windows we'll have to probably ship two `noarch` variants, but worth it to get something not broken-by-default.

.github/workflows/test.yaml
manics:
```suggestion
          pytest -k "not acceptance" -vv
```

.github/workflows/test.yaml
manics: The tests seems to be taking up to 17 minutes, so this is roughly double to allow the occasional slow test run. previously there was no time-limit.

.github/workflows/test.yaml
```diff
@@ -25,19 +25,29 @@ on:
...
+    #       geckowebdriver available. Before upgrading to ubuntu-24.04 in the
+    #       future, check and see its available in a location like this:
+    #       https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers
+    #
```
manics, line 46:
```suggestion
```
I think so since the acceptance tests seem to run OK on Windows. On the other hand I think it's safe to assume Firefox and Geckodriver won't be removed in newer OS images, and if they are we'll find out soon enough

bollwyvl (approved):
:tada:

pyproject.toml
```diff
@@ -47,7 +48,8 @@ dependencies = [
     "aiohttp",
     "importlib_metadata >=4.8.3 ; python_version<\"3.10\"",
     "jupyter-server >=1.0",
-    "simpervisor >=0.4",
+    "simpervisor >=0.4 ; platform_system!=\"Windows\"",
+    "simpervisor >=1.0 ; platform_system==\"Windows\"",
```
consideRatio, line 52: Simpervisor 1.0.0 is not a breaking change for this package that already requires py38, so if we want we can just declare >=1.0 no matter what platform and that should be fine I figure.

bollwyvl: This warms my downstream heart.

consideRatio:
>    # NOTE: The worker environment running this job needs firefox and
    #       geckowebdriver available. Before upgrading to ubuntu-24.04 in the
    #       future, check and see its available in a location like this:
    #       https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers

> On the other hand I think it's safe to assume Firefox and Geckodriver won't be removed in newer OS images, and if they are we'll find out soon enough

It was in software available in 20.04, then removed in 22.04 as the software failed to install/function there I presume, then added in 22.04 in due time. I suspect this will happen again - but hopefully one of us is around to help figure out what broke faster than I did last time, which took quite a while :>

bollwyvl:
> this will happen again

Sure, of course it will break again.

Some options

Instead of a comment, have an explicit pre-flight check that would fail less mysteriously:

```yaml
- name: preflight geckodriver and firefox
  run: |
    echo "Checking for availability of acceptance test dependencies..."
    firefox --version || exit 1
    geckodriver --version || exit 1
    echo "OK"
```

Or split off the acceptance tests as separate jobs, and manage everything with a conda env with e.g. `setup-micromamba`. On ubuntu this _still_ needs some heavyweight stuff (`xvfb`, but usually seems to be around, and `gtk3`), but would at least put a few things under a more managed configuration.

Issue #388, opened 2023-04-08
Release plan for 4.0.0


consideRatio:
Let's plan for the 4.0.0 release!

**Current and proposed breaking changes:**
- #350
- #368
- #387
- #394

**Additional non-blocking nice to get done:**
- #393
- #395 
- Support Windows:
  - Help https://github.com/jupyterhub/simpervisor/pull/26 get merged
  - Help https://github.com/jupyterhub/simpervisor get released
  - Setup tests to run in windows at least once in this repo

yuvipanda:
yes, 100% on keeping jupyter_server 1 support still. I don't have an exact link but somewhere we had to set jupyter_server<2 to preserve compatibility with something else...

manics:
The last release of JupyterLab 2 was in [August 2021](https://pypi.org/project/jupyterlab/2.3.2/). What are the reasons to commit to supporting it for 4.*?

There's discussion in https://github.com/jupyterhub/jupyter-server-proxy/issues/383 about adding new UI components. Without getting into the question of whether we should add new UI components or not, if we do keep JupyterLab 2 it increases the development and testing work required. JupyterLab 4 is also in beta so potentially we'd be supporting three versions soon.

consideRatio:
> The last release of JupyterLab 2 was in [August 2021](https://pypi.org/project/jupyterlab/2.3.2/). What are the reasons to commit to supporting it for 4.*?

With jupyterlab 2 being that old, maybe ~no reason.

👍 For dropping it

manics:
Could we try and get windows support in too? https://github.com/jupyterhub/simpervisor/pull/26

consideRatio:
I added https://github.com/jupyterhub/jupyter-server-proxy/pull/395 to the list of breaking changes as it proposes a change of a API route. I'd love to get that PR part of the release as well no matter what though because it seems like a great feature for this project.

consideRatio:
I figure we shouldn't let #395 block a release, but instead go for faster releases.

@bollwyvl and @manics with #394 in I'll open a changelog PR if you think it makes sense.

consideRatio:
- I also opened #398 that makes sense to merge before a release

consideRatio:
I did `tbump 4.0.0` just a bit too quick, and forgot to refreeze the labextensions yarn.lock file, but cancelled the automation in time to avoid publishing to PyPI/NPM.

I also had to fix the version pattern accepted in `tbump` config.

[...]

PR #394, opened 2023-04-11
maint, breaking: drop support for jupyterlab 2

consideRatio:
This PR is based on top of #393 as I'm modifying the docs also modified there. This PR includes only one commit, [369e5b](https://github.com/jupyterhub/jupyter-server-proxy/pull/394/commits/b369e5bb54105372ded580e84afddc2b7a4e831b).

With this PR, the JupyterLab extension will still be published as an NPM package to @jupyterhub/jupyter-server-proxy, but we will not test against it or ensure future functioning, or document how to install it etc. Dropping support for jupyterlab 2 was proposed to go with the 4.0.0 release planned in(#388 as jupyterlab 4 is coming out and we are to support both jupyterlab 3 and 4 already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Server via the container.
 
 ### Requirements
 
-- `jupyterlab>=2` or `notebook`
+Either `jupyterlab>=3` or `notebook<7` is required.
 
 ### Python package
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ Server via the container.
 
 ### Requirements
 
-- `jupyterlab>=2` or `notebook`
+Either `jupyterlab>=3` or `notebook<7` is required.
```
manics, line 57: Shouldn't notebook>=7 also work, since it's based on JupyterLab?

consideRatio: Maybe hmm maybe, I just saw a lot of tests against notebook<7 specifically as compared to any version of notebook.

Should i adjust this to be as it was, without the <7 part?

bollwyvl:
Can start testing with betas, but otherwise who knows?

manics (approved):

consideRatio:
Going for a merge and a changelog PR to get moving on the release efforts!

Issue #383, opened 2023-03-28
Manager for proxied apps and ability to configure proxied apps CLI args via JupyterLab UI interface


mahendrapaipuri:
### Proposed change

We use jupyter-server-proxy to give access to applications like tensorboard, mlflow, _etc_ to our users _via_ our JupyterHub deployment. These applications (certainly many more) needs CLI args (for instance `logdir` for tensorboard) to be able to work well. This can be tackled by using custom env vars or config files that users need to configure before launching the app. But it would be better if users can set these CLI args directly from the JupyterLab UI before launching the app. 

When user clicks the proxy app on the JupyterLab launcher, it shows an input dialog as below
![labappargs](https://user-images.githubusercontent.com/44365948/228313227-d83f54fb-fc5b-4582-8713-8431899a0ce3.png)
and then user can simply setup CLI args and/or env vars needed to launch that app. We will send these args to the backend, sanitize them and append to the proxy app command. 

Another feature is the ability to launch multiple instances of a given app. This can be achieved by adding one more [config](https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#specifying-config-from-python-packages) option, say `num_instances`, and then registering the handlers with suffix `app_{instance_num}`. This way user can have multiple instances of a given app and it is admins who decide how many instances to allow.

With current implementation of jupyter server proxy, there is no way to terminate the app _via_ UI (user can always send a SIGTERM to process pid "manually"). The idea is to add a manager to the proxied apps that can be managed as kernels, terminals in the JupyterLab sidebar as below
![labapprunning](https://user-images.githubusercontent.com/44365948/228314303-8d98ee53-c38a-44bd-a4f2-9d34bd47c26b.png)
The users can terminate apps here, which will send a SIGTERM signal to the PID that we got from simpervisor.

### Who would use this feature?

All the users that use jupyter server proxy to get UI access to the applications running on a platform. This will be very useful particularly for HPC platforms where JupyterHub/JupyterLab is most common way to provide web access on the compute nodes.

### (Optional): Suggest a solution

We have already implemented these for our platform and it has running well for almost a year. If the developers of jupyter server proxy are interested, we would be happy to open a PR.

ryanlovett:
@mahendrapaipuri Those seem like nice features to me! I would suggest creating separate PRs for each of the new features for discussion.

[...]

Issue #343, opened 2022-07-25
Looking up dynamically allocated port numbers to support testing of proxied services


psychemedia:
Is there a simple pattern for creating basic tests of proxied services, eg using a tool such as `playwright`?

If we know the port number a service is running on, then a simple test of the following form can be used to check that the page loads with the desired title, and then grabs a screenshot of it:

```python
#test_app.py

PORT= 8765

def test_initial_load(page: Page):
    page.goto(f"http://127.0.0.1:{PORT")
    expect(page).to_have_title("Page Title")
    page.screenshot(path="proxied-service-screenshot.png")


#Usage: 
# pip install pytest-playwright
# playwright install
# playwright install-deps
# pytest
```

Using the proxy name alias doesn't seem to get resolved, whereas specifying the port number explicitly does seem to work.

But in a general case, how can we find the port number for a proxied service, particularly if it it allocated dynamically?

I note also that we can get some general metadata about proxied services from `http://localhost/server-proxy/servers-info`, although not the port numbers.

takluyver:
Forgive what's probably a stupid question, but can you test it via the proxy URL (e.g. `http://127.0.0.1:8888/foo`), rather than looking up the port for foo to access it directly?

If JSP launches the server, you need to make at least 1 request via the proxy to ensure it's launched before you try to talk to it. And the server may also know (via some sort of config) what URL it expects to be visible at, so it might not work properly without the proxy URL.

If you do need to access it directly, one option would be to load the same config and entry points that JSP does, and launch services in your test code (at least those using dynamic ports), copying the code to get a random port number:

https://github.com/jupyterhub/jupyter-server-proxy/blob/5ff4c77245a4e92d4c7bb1e7651b65f7a898312b/jupyter_server_proxy/handlers.py#L593-L596

psychemedia:
IIRC, my issue was I have dozens of servers running on different ports, and if I launch a new Jupyter server and to test it, I don't necessarily know what port it's running on if the port is dynamically allocated.

manics:
> Using the proxy name alias doesn't seem to get resolved, whereas specifying the port number explicitly does seem to work.

The proxy name should `server:8888/<service-name>` work with dynamically allocated ports, we have some tests:
https://github.com/jupyterhub/jupyter-server-proxy/blob/31e8b303072d99ddbea9fcdef8169ec19f1f25a4/tests/resources/jupyter_server_config.py#L31-L88
[...]

Issue #264, opened 2021-04-06
Allow users to restart a service


MarcSkovMadsen:
I am and will be enabling users to run Bokeh, Dash, Flask, Panel, Streamlit and Voila apps on Jupyter Hub and Binder.

Those apps are a part of the explorative environment and users would expect to be able to update the underlying code and experience the updated apps. In my case the users might not even update the code them selves but the app code would be updated by other users/ a system team in one or more shared folders that are mounted.

Currently I don't see any way of restarting those services. I would like a feature to easily restart a service.

## Additional Context

Examples of servers are

- [jupyter-panel-proxy](https://github.com/holoviz/jupyter-panel-proxy)
- [custom Panel HighCharts](https://github.com/MarcSkovMadsen/panel-highcharts/blob/main/binder/jupyter-panel-highcharts-server/jupyter_panel_highcharts_server.py)

I have a similar request for `jupyter-panel-proxy` that it should provide functionality to restart. But I also created a feature request here as I believe the use case and request is more general. See https://github.com/holoviz/jupyter-panel-proxy/issues/14

manics:
Would this PR cover your needs? https://github.com/jupyterhub/jupyter-server-proxy/pull/215

bollwyvl:
I think dedicated restarting/status/logging is worth pursuing. Over [on jupyter(lab)-lsp](https://github.com/jupyter-lsp/jupyterlab-lsp/issues/681), we've had similar concerns, especially for language servers that have file-based configurations someone may wish to update interactively.... or that get wedged, and need hard restarts.

While we previously had a dependency on `jupyter-server-proxy`, we ended up needing more bells and whistles, and have been handling our own stuff for a while. We have a [stoppable server session](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/master/python_packages/jupyter_lsp/jupyter_lsp/session.py#L95) concept... but haven't actually ended up exposing it to users. 

Perhaps we could consider some way of bringing some parts of the codebase back together, as a common upstream dependency, for example:

- in python
  - service status/stop handlers
  - logging handlers
- in (retro)lab
  - a streaming/filtered log viewer 
    - text in xtermjs
    - structured logs in datagrid
  - "Running" UIs (with stop, etc)

ideally, these _wouldn't_ specify how the processes actually work: we're using a mishmash of stuff (especially on windows), and might end up having to do some more with e.g. `anyio`

manics:
[...]

PR #395, opened 2023-04-12
Add manager for tracking and terminating running server proxies

mahendrapaipuri:
This PR addresses part of enhancements proposed in #383

**Backend**
Here we are adding a manager on the [backend](https://github.com/jupyterhub/jupyter-server-proxy/commit/085ff479ab755424f61d85c37c9a97dfb53daf54) to keep track of running server proxy apps with assumptions:

- Only managed server proxy apps are added to the manager. Unmanaged apps using #339 will not be added to manager as they cannot be controlled _via_ JupyterLab frontend. 
- Once the simpervisor manages to start the proxy server app successfully, the metadata along with simpervisor process object is added to singleton manager object.
- If the user wishes to terminate a running server proxy app, the API call will effectively send a `SIGTERM` signal to running server proxy app by calling [`terminate`](https://github.com/jupyterhub/simpervisor/blob/ea7466f1feff5780b02807033eb203bb3626cb35/simpervisor/process.py#L165-L175) method on simpervisor process.
- In `ensure_process` method, we are verifying the state of the process using [`proc.running`](https://github.com/jupyterhub/simpervisor/blob/ea7466f1feff5780b02807033eb203bb3626cb35/simpervisor/process.py#L46). When the running server proxy app is either terminated _via_ JupyterLab frontend or if user terminates the app by sending a signal to PID _manually_ `proc.running` will be evaluated to `False`. In these cases, we would like to ideally restart the process again and hence, we remove `proc` object from internal state and restart the process. [_restart_process_if_needed](https://github.com/jupyterhub/simpervisor/blob/ea7466f1feff5780b02807033eb203bb3626cb35/simpervisor/process.py#L122-L137) will make sure simpervisor will restart the process in other cases.

**Frontend**
Manager for the [frontend](https://github.com/jupyterhub/jupyter-server-proxy/commit/6c854787f764bf4783162ce0c4444531dcb63663) follows a similar logic used to implement terminals/kernels [sessions](https://github.com/jupyterlab/jupyterlab/tree/v3.6.3/packages/services/src/terminal) in JupyterLab. 

- Frontend manager makes API requests for every 10s to get the list of running server proxy apps and add them to running sessions in "Server Proxy Apps" section.
- By hovering over an active server proxy app, metadata of the process (command, port, managed) can be viewed.
- User can shutdown a given proxy app or all proxy apps using `shutdown` option in the running sessions. Proxy apps that are shutdown can be restarted afresh from the launcher.

**Ilustration**
ezgif-3-e8dcd463c7
 or all server proxy apps which will terminate by sending `SIGTERM` as explained above.

**Other changes**

- `servers-info` endpoint is moved to `api/server-proxy/servers-info` to be consistent
- `StaticFileHandler` is used for `IconHandler` to simplify code

This extension can be ported to Notebook 7 as well with minimum changes. 

This PR partially addresses #343, #264, #213, #188

@consideRatio @ryanlovett @yuvipanda @bollwyvl Feel free to review and let me know your feedback!!


jupyter_server_proxy/manager.py
```diff
@@ -0,0 +1,127 @@
...
+        for app in self.server_proxy_apps:
+            await self.terminate_server_proxy_app(app)
+
+
+# Create a default manager to keep track of server proxy apps.
+manager = ServerProxyAppManager()
+
+# Create a Periodic call back function to check the status of processes
+pc = PeriodicCallback(monitor_server_proxy_procs, 1e4)
+pc.start()
```
consideRatio, line 127: This could be reasonable, but I'm not confident on it. Is it acceptable that we start this up automatically as soon as this file is loaded?

I'm not confident on the alternate options, but I figure this is a singleton instance of the class defined above, created when this python file is loaded. I figure options involve delaying this initialization.

I'm also thinking a bit about if `monitor_server_proxy_procs` should or shouldn't be part of the class itself. It is a function tied to a specific object of the class anyhow via the `manager` variable that could be `self` as well.

If not someone has a concrete change idea here to persue or generally considers this to be blocking, I don't think this comment should be blocking a merge.

consideRatio (changes requested):
@mahendrapaipuri this looks great to me overall!!! :heart: :tada: :sunflower:

I see a clear value in this feature, it also seems like a well thought out design and implementation of the design! I've not been able to review the JS parts thoroughly as I'm not experienced with that, but from what I can tell it looks great as well.

I appreciate a lot about this PR but I'd like to highlight that your PR description was amazing to help me review it, I also greatly appreciate that it included references to related issues.

### Required work towards merge

Before doing anything to avoid merge conflicts etc, please __rebase on the current main branch__ to acquire #393 and #396 that changes things in docs and testing infra.

- [ ] Update docs to reflect that the labextension does more than just add launcher buttons, but that it also provides these new UI to manage servers
- [ ] Add entry in the changelog under the unreleased title that the api endpoint has been renamed
- [ ] Consider if we should print some log messages someplace.

### Optional bonus work

I note that we don't have existing tests of for example the `servers-info` endpoint. Since its become a far bigger project to setup tests from scratch than to mimic existing tests, I don't think we should block merging of this PR as it is but instead open an issue to represent the need to develop tests of our backend API separately going onwards.

- [ ] Add tests of introduced backend API handlers
  - [ ] GET `api/server-proxy` (to list)
  - [ ] GET `api/server-proxy/<proxy name>`, failure situation + success situation
  - [ ] DELETE `api/server-proxy/<proxy name>`, failure situation + success situation
- [ ] Add browser / acceptance test of this feature

jupyter_server_proxy/__init__.py
```diff
@@ -54,21 +59,29 @@ def _load_jupyter_server_extension(nbapp):
...
+                ujoin(base_url, "api/server-proxy/servers-info"),
                 ServersInfoHandler,
                 {"server_processes": server_processes},
             ),
-            (ujoin(base_url, "server-proxy/icon/(.*)"), IconHandler, {"icons": icons}),
-        ],
+            (ujoin(base_url, r"api/server-proxy"), ListServersAPIHandler),
+            (ujoin(base_url, r"api/server-proxy/(?P<name>.*)"), ServersAPIHandler),
+        ]
+        + icon_handlers,
```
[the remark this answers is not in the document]
bollwyvl: Welp, declaring (and testing against) an OpenAPI spec would be a good place to start, so that one doesn't have to _read the code_ to figure out what's going on. Prior art:
- https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/services/api/api.yaml
- https://github.com/jupyterlab/jupyterlab_server/blob/main/jupyterlab_server/rest-api.yml

This can generate nice static docs, as well as help cleaning up some of the design decisions: if you can't express a pattern as a URL param or in a schema, maybe it's not a great idea.

It also then becomes possible to re-use that schema declaration to generate lightweight python and TS types, even if one doesn't do fully generated clients.

jupyter_server_proxy/manager.py
bollwyvl: Right. Importing this file shouldn't have side-effects like starting/changing processes or even things on the event loop, as there's no telling it will actually be there.

Ideally, as much business logic/exception handling as possible would move into the manager class, and out of the handlers and `__init__.py`. The server extension would then become a very small layer that just hands down the parent application into the manager.

bollwyvl: Also, I'd recommend new-start async stuff to use stdlib `asyncio` or  [`anyio`](https://anyio.readthedocs.io/en/stable/) primitives instead of `tornado`-specific ones. And indeed, if we could get _out_ of the polling business, that would be ideal: the underlying processes probably do (or should) provide callbacks/events that can be used to directly update the manager state.

Assuming we _do_ have to poll, for some reason, and as I don't actually see a drop-in for `PeriodicCallback`, this could be a `ServerProxyManager.monitor()` which fires off an infinite loop:

```python
import traitlets
import asyncio 

class ServerProxyManager(ServerProxy):
    monitor_interval = traitlets.Int(10, help="Proxy polling interval in seconds").tag(config=True)
    monitoring = traitlets.Bool(False, help="Whether the manager should poll for proxy status")

    # this doesn't need to be a trait
    _monitor_task: asyncio.Task

    @traitlets.observe("monitoring")
    def on_monitoring_changed(self, change: traitlets.Bunch):
        if self._monitor_task:
            self._monitoring_task.cancel()
        if change.new:
            # keep a handle so it doesn't get GCd
            self._monitoring_task = asyncio.create_task(self._monitor_loop()) 

    async def _monitor_loop(self):
        while self.monitoring:
             # do the thing
             await asyncio.sleep(self.monitor_interval)
```

When this gets invoked from `__init__.py`:

```
def _load_jupyter_server_extension(nbapp: ServerApp):
    manager = ServerProxyManager(parent=nbapp)
    manager.monitor()
```

labextension/src/index.ts
```diff
@@ -32,6 +46,54 @@ function newServerProxyWidget(
...
+        (model) => new RunningServerProxyApp(model),
+      ),
+    shutdownAll: () => manager.shutdownAll(),
+    refreshRunning: () => manager.refreshRunning(),
+    runningChanged: manager.runningChanged,
+    shutdownAllConfirmationText:
+      "Are you sure you want to close all server proxy applications?",
+  });
+
+  class RunningServerProxyApp implements IRunningSessions.IRunningItem {
```
bollwyvl, line 73: This should be a separate file.

labextension/src/index.ts
```diff
@@ -105,6 +175,15 @@ async function activate(
     },
   });
 
+  commands.addCommand(CommandIDs.open, {
+    execute: (args) => {
```
bollwyvl, line 179: recommend typing the args beyond `any`, and making them more descriptive.

labextension/src/index.ts
```diff
@@ -105,6 +175,15 @@ async function activate(
...
+      const model = args["sp"] as IServerProxyModel;
+      const url = PageConfig.getBaseUrl() + model.url;
```
bollwyvl, line 181: Generally, one should use the `URLExt.join` feature instead of string concatenation/templating when working with URLs.

labextension/src/serverproxy.ts
```diff
@@ -0,0 +1,90 @@
...
+  managed: boolean;
+}
+
+/**
+ * The interface for a server proxy manager.
+ *
+ * The manager is responsible for maintaining the state of running
+ * server proxy apps.
+ */
+export interface IManager {
```
bollwyvl, line 41: This can `extend` other interfaces so one needn't retype e.g. `shutdownAll`.

labextension/src/restapi.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+export async function shutdown(
+  name: string,
+  settings: ServerConnection.ISettings = ServerConnection.makeSettings(),
+): Promise<void> {
+  const url = URLExt.join(settings.baseUrl, SERVER_PROXY_SERVICE_URL, name);
+  const init = { method: "DELETE" };
+  const response = await ServerConnection.makeRequest(url, init, settings);
+  if (response.status === 404) {
+    const msg = `Server proxy "${name}" does not exist. Are you sure "${name}" is started by jupyter-server-proxy?`;
```
bollwyvl, line 52: New-start, complex strings should using the JupyterLab translation capabilities. Something like this should probably be hoisted to someplace other than the console (or an on-screen cue should indicate there is more in the console).

labextension/src/restapi.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * List the running server proxy apps.
+ *
+ * @param settings - The server settings to use.
+ *
+ * @returns A promise that resolves with the list of running session models.
+ */
+export async function listRunning(
+  settings: ServerConnection.ISettings = ServerConnection.makeSettings(),
```
bollwyvl, line 18: having the manager do this (with exactly one `settings`, handed in at startup) would make these signatures a lot simpler.

labextension/src/manager.ts
```diff
@@ -0,0 +1,164 @@
...
+   */
+  constructor(options: ServerProxyManager.IOptions = {}) {
+    this.serverSettings =
+      options.serverSettings || ServerConnection.makeSettings();
+    this._refreshTimer = (setInterval as any)(() => {
+      if (typeof document !== "undefined" && document.hidden) {
+        return;
+      }
+      this._refreshRunning();
+    }, 10000);
```
bollwyvl, line 21: this should likely be configurable

labextension/src/index.ts
```diff
@@ -5,7 +5,21 @@ import {
...
+import { IModel as IServerProxyModel } from "./serverproxy";
+import serverProxyAppSvgstr from "../style/icons/proxy.svg";
+
+export const ServerProxyAppIcon = new LabIcon({
+  name: "server-proxy:proxyAppIcon",
+  svgstr: serverProxyAppSvgstr,
+});
+
+namespace CommandIDs {
+  export const open = "running-server-proxy:open";
```
bollwyvl, line 21: Probably need a broader command area surface, e.g. `server-proxy:refresh`

bollwyvl:
This is a reasonable starting point, but would benefit from aligning with the configuration and async patterns used in the upstream `jupyter_server`, as well as siblings such as `jupyterlab_server` and [`jupyter_lsp`](https://github.com/jupyter-lsp/jupyterlab-lsp/blob/master/python_packages/jupyter_lsp/jupyter_lsp/manager.py).

I'd say a lack of _any_ python tests for the new features is probably a hard blocker to this getting merged, but really, with the more complex lab stuff, this would need to be tested by robot tests as well to be seriously considered for merging.

As noted, in general when adding a new REST APIs, starting with the schema and generating the python and TS types is a really good way to keep the system manageable, over time, and whether built with OpenAPI or not, can make the documentation process also much smoother.

jupyter_server_proxy/manager.py
```diff
@@ -0,0 +1,127 @@
+"""Manager for jupyter server proxy"""
+
+from collections import namedtuple
+
+from jupyter_server.utils import url_path_join as ujoin
+from tornado.ioloop import PeriodicCallback
+
+from .utils import check_pid
+
+ServerProxy = namedtuple("ServerProxy", ["name", "url", "cmd", "port", "managed"])
```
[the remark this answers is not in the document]
mahendrapaipuri: Yes, it is pretty straightforward. Will do it!!

jupyter_server_proxy/__init__.py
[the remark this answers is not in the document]
mahendrapaipuri: Agree with OpenAPI suggestion. That would make it more cleaner. Actually, `servers-info` endpoint is only handling ["launcher" specific metadata](https://github.com/jupyterhub/jupyter-server-proxy/blob/407a1a4809c598dd30647a000343a2f32a145f1e/jupyter_server_proxy/api.py#L19-L30) whereas the new endpoints are handling [server specific metadata](https://github.com/jupyterhub/jupyter-server-proxy/blob/bac7ccce3bf84c529eeb78a72fcd85bbd8892490/jupyter_server_proxy/api.py#L76). So, I think it is logical to keep them as separate. However, I think we need to change the name of `servers-info` endpoint as it is bit misleading? `ServersAPIHandler` and `ListServersAPIHandler` can be merged into one using query parameters.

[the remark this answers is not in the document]
mahendrapaipuri: @bollwyvl Cheers for the refs

jupyter_server_proxy/manager.py
mahendrapaipuri: If a user gracefully terminates a server proxy _manually_ outside of lab interface, that process will not be restarted as we are not using [`always_restart`](https://github.com/jupyterhub/simpervisor/blob/425c81f56c76756e3a02cb9f9a108d543964324d/simpervisor/process.py#L134) feature of simpervisor. So somehow we need to update manager to remove this server proxy app and that is why I included this polling. Simpervisor uses exit handlers internally but I think they are not configurable. So, I dont know how we can fire a callback when process exits gracefully.

mahendrapaipuri:
@consideRatio @bollwyvl Thanks for the feedback. I will work on the changes. Yes, I agree that we need to add tests for these features and I am working on them already. Next week I will be at KubeCon so, I might not find lot of time to work on this.

Are you aiming to make this extension compatible with Notebook 7? Currently the frontend extension fails as it is loading JupyterLab specific components. Straightforward solution is to load UI components conditionally, _i.e.,_ making [`ILauncher`](https://github.com/jupyterhub/jupyter-server-proxy/blob/bac7ccce3bf84c529eeb78a72fcd85bbd8892490/labextension/src/index.ts#L224) optional. Adding an optional Notebook related component (that add servers to Notebook launcher) to the extension should make it work for both Lab and Notebook interfaces. @bollwyvl Is it the strategy being followed now? I could not find a lot of info about extensions in new Notebook 7?

jupyter_server_proxy/utils.py
```diff
@@ -49,3 +51,13 @@ def validate(self, obj, value):
             return value
         else:
             self.error(obj, value)
+
+
+def check_pid(pid):
+    """Check For the existence of a unix pid"""
```
manics, line 57: We're hoping to add Windows support soon to jupyer-server-proxy. Is this unix specific, or should it work on Windows too?

mahendrapaipuri:
@consideRatio @bollwyvl @manics I have update the PR with most of the comments addressed. Please check the updated PR description for more details.

On a side note, I think I did something strange with rebasing. I am not quite sure why commits from previous PRs are showing up here. On my VSCode I have a proper commit history though!! Any help will be appreciated!! Cheers!!

jupyter_server_proxy/utils.py
mahendrapaipuri: This is not needed anymore. We can check this directly on the simpervisor proc object. If simpervisor supports Windows, that should be working here as well!!

consideRatio: Seen to be resolved by d1653d11babfd9717a5b81a2e42e3031d3c79462

consideRatio:
This is amazing work @mahendrapaipuri!!!

I'm truly sorry we've not been able to review your work further yet @mahendrapaipuri. I'll do a small push now but I can't review this myself to completion.

As this PR is very large, avoid force pushing to the PR branch to help reviewers keep track of changes since last review better. I'll go through a few comments and link to the resolving commit to help other reviewers. I'd like to get the open review comments down to a minimum to reduce complexity of further review.

jupyter_server_proxy/manager.py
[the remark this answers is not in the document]
consideRatio: Addressed by d1653d11babfd9717a5b81a2e42e3031d3c79462 I think

jupyter_server_proxy/handlers.py
```diff
@@ -836,6 +844,12 @@ async def ensure_process(self):
                     if not is_ready:
                         await proc.kill()
                         raise web.HTTPError(500, f"could not start {self.name} in time")
+
+                    # If process started succesfully, add it to manager
+                    # Add the server proxy app to manager
```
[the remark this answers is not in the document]
consideRatio: No comments around about this any more - marking as resolved

jupyter_server_proxy/api.py
```diff
@@ -34,39 +36,54 @@ async def get(self):
         self.write({"server_processes": data})
 
 
-# FIXME: Should be a StaticFileHandler subclass
-class IconHandler(JupyterHandler):
-    """
-    Serve launcher icons
-    """
+# Took it from JupyterHub LogoHandler
```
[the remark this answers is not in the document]
consideRatio: This change is included now

labextension/src/index.ts
consideRatio: This class relocated to running.ts

labextension/src/index.ts
consideRatio: Observed to be resolved now

labextension/src/restapi.ts
consideRatio: Looking at the changes as they are now, i see this - does that make this comment resolved?

```typescript
  if (response.status === 404) {
    const msg = trans.__(
      `Server proxy "${name}" is not running anymore. It will be removed from this list shortly`,
    );
    console.warn(msg);
```

labextension/src/manager.ts
consideRatio: It now is

```
    monitor_interval = Int(
        10,
        help="""
        Proxy polling interval in seconds. The server proxy manager will keep
        polling the status of running servers with a frequency set by this
        interval.
        """,
        config=True,
    )
```

mahendrapaipuri:
Thanks a lot @consideRatio for reviewing the changes. It is totally fine, we all are busy with a lot of stuff and I understand things can take time. 

I guess there are still few small things to address and I will do it next week. Cheers!!

labextension/src/index.ts
mahendrapaipuri: This is done in #425 and once we rebase onto `main` it will be addressed

labextension/src/serverproxy.ts
mahendrapaipuri: `ServerProxyManager` is now extended from `BaseManager` in `@jupyterlab/services`. Ref [commit](https://github.com/jupyterhub/jupyter-server-proxy/pull/395/commits/f25996b57d452ecfdd85fb5bf02b0f6c6ef147ae)

labextension/src/restapi.ts
mahendrapaipuri: Yes. Also, now we show a modal dialog with that message.

jhgoebbert:
This adds great additional features to `jupyter-server-proxy`. I am looking forward to see it merged.

PR #398, opened 2023-04-17
maint: update package description to not claim windows support

consideRatio:
Our package is described to support use with windows, but we are not there yet. Pending work to support this, we remove the description that we support it.

### Related

- #147 
- #181 
- #392 
- https://github.com/jupyterhub/simpervisor/pull/26


yuvipanda:
We can add it back if it works!

PR #419, opened 2023-07-28
Add JupyterLab 4 support for the extension

trungleduc:
Add JupyterLab 4 support for the extension.


labextension/.yarnrc.yml
```diff
@@ -0,0 +1,3 @@
+nodeLinker: node-modules
```
manics, line 1: What's the purpose of this new file?

trungleduc: This is the yarn 3 configuration file. If you develop the extension against jupyterlab 4, `jlpm` is actually yarn 3 instead of yarn 1 in the case of lab 3

manics: Thanks! Does every extension repo need this config file, or is it a property of the development environment? Does the package registry have to be hardcoded, or can we rely on the developers default e.g. if they're using an internal mirror?

trungleduc: > Thanks! Does every extension repo need this config file

Yes

> Does the package registry have to be hardcoded, or can we rely on the developers default e.g. if they're using an internal mirror?

No, it should not be hard-coded, I only convert the entries in the old config file (https://github.com/jupyterhub/jupyter-server-proxy/blob/main/labextension/.yarnrc). For other extensions only the `nodeLinker` config is required.

bollwyvl:
If this is worth doing (and it probably is) it should also support Notebook 7, to the best of its abilities. As NB7 doesn't have a launcher, it would need to be ported to a file browser dropdown to stay consistent, UX-wise.

There is some work on-going to update documentation for supporting lab4/nb7 in the same extension, which can likely be used here: basically checking whether `ILabShell` is available should be enough.

Much as I tried (admittedly, not very hard), I ended up not being able to cleanly do keyword-for-keyword compatibility between jquery- and lumino-based notebooks, so the browser-based acceptance tests would need to updated to use the [new client keywords](https://robotframework-jupyterlibrary.readthedocs.io/en/stable/KEYWORDS.html#jupyter-notebook).

trungleduc:
Indeed It would be nice to support both Lab 4 and Notebook 7, but I think It can be addressed in a separate PR.

trungleduc:
Hi @ryanlovett  @takluyver @consideRatio @yuvipanda, can someone take a look at this PR? 
Thanks!

ryanlovett:
Hi @trungleduc, I do think it'd be wonderful to support JL4, but I'm afraid the javascript bits are not my strong suit.

To me it seems fine to add support for Notebook 7 in a separate PR given that it has a different launch mechanism. I imagine this could be the `New` menu in the file browser as it was in previous notebook versions, or something new.

Khagesh16:
Really appreciate jlab 4 upgrade

labextension/.yarnrc.yml
consideRatio, line 1: I think this PR is ready to go, but it would be helpful to have the insights from this conversation captured in inline comments, combined with a configuration reference link or so to this kind of file in order to help us maintain this long term.

I may come back soon and add a code suggestion -> apply it -> and merge.

For inspiration on a top level comment, look at for example pre-commit-config.yaml that introduces the file and helps the reader understand how to maintain things in it a better.

trungleduc: This file is a standard configuration in all other Jupyter projects: [JupyterLab](https://github.com/jupyterlab/jupyterlab/blob/main/.yarnrc.yml), [Notebook](https://github.com/jupyter/notebook/blob/main/.yarnrc.yml), [ipywidgets](https://github.com/jupyter-widgets/ipywidgets/blob/main/.yarnrc.yml) and also in the [template of JupyterLab extensions](https://github.com/jupyterlab/extension-cookiecutter-ts/blob/4.0/%7B%7Bcookiecutter.python_name%7D%7D/.yarnrc.yml)

consideRatio:
- I pushed a commit commenting yarn configs and fixed details in a README file
- Tests failures are unrelated, tracked in https://github.com/jupyterhub/jupyter-server-proxy/issues/421.
- I opened #420 to track additional work to make this software provide a Notebook 7 compatible frontend extension as well.

Thank you for working this @trungleduc!!

consideRatio (approved):

PR #423, opened 2023-09-19
Add changelog for 4.1.0

consideRatio:
I think its worth getting a release out for #419 that makes the jupyterlab extension work in jupyterlab 4 environments, and looking at the related bugfix #415, it also seems quite important for users of the unix_socket feature.

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## 4.1
+
+### 4.1.0 - 2023-09-25
+
+#### New features added
+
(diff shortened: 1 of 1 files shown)


yuvipanda (approved):
LGTM! Merge when you feel appropriate. Thanks for doing this work!

consideRatio:
@bollwyvl I've updated this to include your recent PR now as well, I think we should go straight for 4.1.0 - now. Wdyt?

consideRatio:
Wiee with a :+1: from @bollwyvl and approval from @yuvipanda I'll go for a self-merge and release!

takluyver:
Belatedly, thanks for making this release! :tada:

Issue #406, opened 2023-04-28
Use pytest-jupyter fixtures in unit testing


mahendrapaipuri:
### Proposed change
In a recent PR #396, unit testing is changed from running a external JupyterLab server to running the server via fixtures. This is already a good contribution as we do not need to run anything manually to do testing. The downside is that tests take way longer now as each test is spinning up a jupyter server (I guess?)

Aren't we supposed to use [pytest-jupyter](https://pytest-jupyter.readthedocs.io/en/main/index.html) fixtures for this use cases? We have [`jp_fetch`](https://github.com/jupyter-server/pytest-jupyter/blob/aba6e6c8773ec99a338d277d22a1c39fc584ab03/pytest_jupyter/jupyter_server.py#L290-L319) fixture which is what we need to test our cases. There is also sibling [`jp_ws_fetch`](https://github.com/jupyter-server/pytest-jupyter/blob/aba6e6c8773ec99a338d277d22a1c39fc584ab03/pytest_jupyter/jupyter_server.py#L323-L366) for WS connection. We can use them "out-of-the-box" in the unit tests. I have tested them locally , they work fine and they are way faster. Is there any specific reason why these fixtures are not being used?

If not, I can put up a PR.

### Who would use this feature?
Anyone who is contributing to the JSP project.

PR #425, opened 2023-09-19
Add extension for Jupyter Notebook 7

### LINKED ISSUE #420: JupyterLab extension to also be a Notebook 7 extension
From @bollwyvl in https://github.com/jupyterhub/jupyter-server-proxy/pull/419#issuecomment-1677987290

> If this is worth doing (and it probably is) it should also support Notebook 7, to the best of its abilities. As NB7 doesn't have a launcher, it would need to be ported to a file browser dropdown to stay consistent, UX-wise.
> 
> There is some work on-going to update documentation for supporting lab4/nb7 in the same extension, which can likely be used here: basically checking whether `ILabShell` is available should be enough.
> 
> Much as I tried (admittedly, not very hard), I ended up not being able to cleanly do keyword-for-keyword compatibility between jquery- and lumino-based notebooks, so the browser-based acceptance tests would need to updated to use the [new client keywords](https://robotframework-jupyterlibrary.readthedocs.io/en/stable/KEYWORDS.html#jupyter-notebook).

### LINKED ISSUE #424: Do we need .yarnrc, or can we always build the extension with jupyterlab 4 that doesn't need it?
I understand that we have `.yarnrc` to enable jupyterlab <4 to build the jupyterlab extension, and `.yarnrc.yml` to enable jupyterlab>=4 to build the jupyterlab extension.

But since we seem to have a pyproject.toml that declares build env stuff, would it be sufficient for us to update this part to require jupyterlab 4 in there and build the extension with jupyterlab 4 also for jupyterlab 3?

https://github.com/jupyterhub/jupyter-server-proxy/blob/7360d846f1786db4419bac7df3671cb72acb0ba8/pyproject.toml#L1-L12

bollwyvl:
## References

- fixes #420
- fixes #424
- partially addresses some motivations on #406

## Changes
- [x] UI
  - [x] add dropdown to tree button bar for parity with nb6 (https://github.com/jupyterhub/jupyter-server-proxy/pull/419#issuecomment-1686683413)
- [x] python
  - [x] update dependency ranges
- [x] js/ts
  - [x] handle missing `ILauncher`
  - [x] extract a number of constants and types to `tokens.ts`
  - [x] adds a local `prettier` install to account for CI updates
    - > pre-commit won't do the PR update if you mess with CI
  - [x] update `typescript`
- [x] testing
  - [x] use `session` scoped server fixture
  - [x] increase
- [x] CI
  - [x] build once in CI, download into others
  - [x] run full `pytest` suite in single step
    - > acceptance test coverage is then included in the HTML coverage report

## Screens

a screenshot of the notebook 7 interface

## Breaking Changes

- shouldn't change the signature of anything


labextension/.yarnrc.yml
```diff
@@ -1,8 +1,23 @@
...
+  - code: YN0002
+    level: discard
+  - code: YN0007
+    level: discard
+  - code: YN0013
+    level: discard
+  - code: YN0019
+    level: discard
+  - code: YN0008
+    level: discard
```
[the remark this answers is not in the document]
bollwyvl: updated in 1842173

.github/workflows/test.yaml
```diff
@@ -128,28 +130,37 @@ jobs:
...
-          jupyter labextension list 2>&1 | grep -iE '@jupyterhub/jupyter-server-proxy.*OK.*'
+          jupyter lab extension list
+          jupyter lab extension list 2>&1 | grep -iE 'jupyter_server_proxy.*OK.*'
 
       - name: Run tests
         run: |
-          mkdir build
-          cd build
-          pytest -vv ../tests -c ../pyproject.toml
+          pytest
```
bollwyvl, line 163: having a look at the coverage, this change results in finding the "at rest" `jupyter_server_proxy` folder, _not_ the as-installed one, during tests. while not a _huge_ deal, it screws up a bunch of the metrics.

while it's annoying to get coverage out to the github issue itself without tokens, we _can_ make a markdown preview and write it out to the ci summary.

some ways around this 
- `rm -rf jupyter_server_proxy` before invoking `pytest`
- adopt the `src/jupyter_server_proxy` layout

bollwyvl: example: https://github.com/jupyterhub/jupyter-server-proxy/actions/runs/6263255269/job/17007322266?pr=425#step:13:111

consideRatio: Yikes ooops I pushed to this branch by mistake =/

I meant to try if this worked still in a private branch in my fork because I didn't understand it why it was done.

What do you think about not doing this tested change and just adding a comment?

tests/acceptance/test_acceptance.py
```diff
@@ -22,6 +29,12 @@ def test_robot():
     pytest.importorskip("JupyterLibrary")
 
     env = dict(**os.environ)
+    robot_args = ["robot", "--consolecolors=on", f"--outputdir={OUTPUT}"]
+
+    if NOTEBOOK_VERSION >= 7:
```
consideRatio, line 34: This variable could be None, then we would run into `TypeError: '>=' not supported between instances of 'NoneType' and 'int'`.

```suggestion
    if NOTEBOOK_VERSION and NOTEBOOK_VERSION >= 7:
```

labextension/package.json
```diff
@@ -40,18 +40,21 @@
...
     "watch:labextension": "jupyter labextension watch .",
-    "deduplicate": "yarn-deduplicate -s fewer --fail"
+    "deduplicate": "yarn-deduplicate -s fewer --fail",
+    "lint:prettier": "prettier --write \"../.github/**/*.yaml\" \"../*.{yaml,yml,md}\" \"../docs/**/*.md\" \"src/**/*.{tsx,ts}\" \"./*.{yml,json,md}\""
   },
   "dependencies": {
-    "@jupyterlab/application": "^2.0 || ^3.0 || ^4.0",
-    "@jupyterlab/launcher": "^2.0 || ^3.0 || ^4.0"
+    "@jupyterlab/application": "^3.0 || ^4.0",
+    "@jupyterlab/filebrowser": "^3.0 || ^4.0",
```
consideRatio, line 48: We have stopped testing jupyterlab 2 according to 4.0.0 changelog, and this is now also removing support for it right?

consideRatio (commented):
I think this looks good from the technical perspective, but I'm not confident at all on the extension code =/

There is some documentation that is worth updating as part of this I think. We have a "labextesion" folder, but its now an extension for JupyterLab 3-4, and an extension for Notebook 7? I didn't know about it, but there was some behavior in notebook 6 also?

As you probably overview the typescript code base quite well at this point, could you provide some code contributor / maintainer facing docs overviewing what the various code files does?

There is a labextensio/README.md that could be a place to write some comments as well on how it works. That file is also outdated currently I think, as there is a claim about jupyterlab >=2,<5 is supported.

.github/workflows/test.yaml
consideRatio: I've reverted that commit and added a comment as I understand it

labextension/package.json
consideRatio: Ah, nevermind, we said:

> jupyter-server-proxy's JupyterLab extension dropped support for JupyterLab 2 (https://github.com/jupyterhub/jupyter-server-proxy/pull/394)

Then we have some cleanup in other places about this to do.

consideRatio (changes requested):
Thank you for working this @bollwyvl!!! :heart: :tada: :sunflower: 

I've struggled a bit since this is not a domain I have much experience in, but I have some suggestions related to docs to a large extent.

- [ ] Resolve comment about `NOTEBOOK_VERSION`
- [ ] Revise README.md's heading `### JupyterLab extension`
- [ ] Revise labextension/README.md
  - Is this a Notebook 7 + JupyterLab extension now?
  - Is it a Notebook 6 extension somehow also? I was surprised about the notebook 7 functionality added already was found in notebook 6.
- [ ] (optional) Add top-level comments for .ts files providing a high level overview in a paragraph or two
  - Something like this will help me greatly as a novice about code for frontend extensions in reviewing or possibly fixing a bug etc.

consideRatio:
I've observed a few address already in use failures, I'm not sure if they have started showing up more and more with this PR or not, but they are quite frequent now. I've seen them about ~50% of the times the tests run.

Could it be related to making fixtures session based?

bollwyvl:
> Revise labextension/README.md

This populates the entry on `npmjs.org`, but as this package hasn't been published under its current name, it's... not. I've updated the main `README` with a compatibility matrix, which is likely more relevant.

> Is this a Notebook 7 + JupyterLab extension now?

Yes, that's the point of this PR. Notebook 7 is built of 95% jupyterlab components, so mostly everything works the same, but you have to watch out for `notebookPage`.

> Is it a Notebook 6 extension somehow also? I was surprised about the notebook 7 functionality added already was found in notebook 6.

Yes, the existing `tree.js` still does the thing. I, for one, would not sign up to add new features there, as `notebook<7`, `nbclassic` is _really_ on life support at this point, with basically no non-emergency maintainers. But as it's only two files, there's no particular harm in keeping it, as it doesn't force another build.

tests/acceptance/test_acceptance.py
bollwyvl: 8f0ff5e

Turns out we _could_ also test the classic stuff by using the `nbclassic` shim, but... meh.

consideRatio (approved):
This looks good to me, and with the amazing work with browser tests I think we can go for a merge even though I'm not so confident reviewing the frontend code. What do you think @bollwyvl?

bollwyvl:
> address already in use failures

This one's tricky. Getting rid of the hard-coded port (e.g. `NOT_54321_PORT = get_unused_port()`) would _for sure_ get rid of this, but it might be indicative of a deeper cleanup problem... perhaps the `atexitasync` stuff isn't firing properly, or some other `async` shenanigans? Perhaps it's not interacting correctly with `pytest`'s freaky stuff?

bollwyvl:
>  go for a merge 

Sure, why not. Could do an `4.1.0a0` or something to try it out, but we've done the best we can to _not_ break anything, and sometimes only peoples' unpinned prod environments find the bugs.

consideRatio:
I opened https://github.com/jupyterhub/jupyter-server-proxy/issues/426 about the intermittent test failrues, lets go for it!

PR #427, opened 2023-09-25
Declare metadata in pyproject.toml, remove hatch-nodejs-version plugin

### LINKED ISSUE #429: Test suite is failing in main branch
This relates to https://github.com/jupyterhub/jupyter-server-proxy/pull/427 and me pushing a version (valid semver2 version, but not accepted).

consideRatio:
When I made a release I bumped to a dev version after, but that had `pyproject-build` start to fail. This was because the `hatch-nodejs-version` plugin didn't allow us to use a version in package.json that was `4.1.1-0.dev` which is a valid semver2 version. In helm chart's etc that require semver2 compliant versions like package.json also wants, we've ended up using the `-0.dev` version suffix as it allows us to sort that version before alpha if we make something like that sometime.

In this PR I've instead of adjusting the version 4.1.1-0.dev to something that worked with `hatch-nodejs-version` plugin adjusted to no longer use it. I've as part of this removed some metadata from package.json and put it directly in pyproject.toml. The version field wasn't put directly there, but in `_version.py`, allowing it to be inspected via an import statement by other packages.

- Closes #429

### DIFF
--- a/jupyter_server_proxy/__init__.py
+++ b/jupyter_server_proxy/__init__.py
@@ -1,5 +1,6 @@
 from jupyter_server.utils import url_path_join as ujoin
 
+from ._version import __version__  # noqa
 from .api import IconHandler, ServersInfoHandler
 from .config import ServerProxy as ServerProxyConfig
 from .config import get_entrypoint_server_processes, make_handlers, make_server_process
--- a/jupyter_server_proxy/_version.py
+++ b/jupyter_server_proxy/_version.py
@@ -1,0 +1,4 @@
+# __version__ should be updated using tbump, based on configuration in
+# pyproject.toml, according to instructions in RELEASE.md.
+#
+__version__ = "4.1.1-0.dev"
--- a/labextension/package.json
+++ b/labextension/package.json
@@ -1,14 +1,31 @@
 {
   "name": "@jupyterhub/jupyter-server-proxy",
   "version": "4.1.1-0.dev",
+  "description": "A JupyterLab extension accompanying the PyPI package jupyter-server-proxy adding launcher items for configured server processes.",
+  "keywords": [
+    "jupyter",
+    "jupyterlab",
+    "jupyterlab-extension"
+  ],
   "homepage": "https://github.com/jupyterhub/jupyter-server-proxy",
+  "bugs": {
+    "url": "https://github.com/jupyterhub/jupyter-server-proxy/issues"
+  },
   "license": "BSD-3-Clause",
+  "author": {
+    "name": "Ryan Lovett & Yuvi Panda",
+    "email": "rylo@berkeley.edu"
+  },
   "files": [
     "LICENSE",
     "lib/**/*.{d.ts,eot,gif,html,jpg,js,js.map,json,png,svg,woff2,ttf}"
   ],
   "main": "lib/index.js",
   "types": "lib/index.d.ts",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/jupyterhub/jupyter-server-proxy.git"
+  },
   "scripts": {
     "build": "jlpm run build:lib && jlpm run build:labextension:dev",
     "build:prod": "jlpm clean && jlpm run build:lib && jlpm run build:labextension",
--- a/labextension/package.json
+++ b/labextension/package.json
@@ -1,31 +1,14 @@
 {
   "name": "@jupyterhub/jupyter-server-proxy",
   "version": "4.1.1-0.dev",
-  "description": "A JupyterLab extension accompanying the PyPI package jupyter-server-proxy adding launcher items for configured server processes.",
-  "keywords": [
-    "jupyter",
-    "jupyterlab",
-    "jupyterlab-extension"
-  ],
   "homepage": "https://github.com/jupyterhub/jupyter-server-proxy",
-  "bugs": {
-    "url": "https://github.com/jupyterhub/jupyter-server-proxy/issues"
-  },
   "license": "BSD-3-Clause",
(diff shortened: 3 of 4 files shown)


jupyter_server_proxy/_version.py
```diff
@@ -0,0 +1,4 @@
+# __version__ should be updated using tbump, based on configuration in
+# pyproject.toml, according to instructions in RELEASE.md.
+#
+__version__ = "4.1.1-0.dev"
```
bollwyvl, line 4: tbump is fine and all, but it could dispense with this copy with e.g.

```py
__version__ == __import__("importlib.metadata").metadata.version("jupyter_server_proxy")
```

labextension/package.json
```diff
@@ -1,31 +1,14 @@
 {
   "name": "@jupyterhub/jupyter-server-proxy",
   "version": "4.1.1-0.dev",
```
bollwyvl, line 3: If the labextension was published to `@jupyterhub/jupyter-server-proxy` on npmjs (e.g. if it exposed some extension point for the features in #395) then the README page would look pretty bad indeed. Recommend keeping these and/or generating them from `pyproject.toml` as the single source of truth.

npm and its raft of things has a different schema for valid versions, which would also need to be take into account.

pyproject.toml
```diff
@@ -17,13 +16,14 @@ requires = [
 #
 [project]
 name = "jupyter_server_proxy"
-dynamic = [
-    "authors",
-    "description",
-    "keywords",
-    "urls",
-    "version",
+description = "A JupyterLab extension accompanying the PyPI package jupyter-server-proxy adding launcher items for configured server processes."
```
bollwyvl, line 19: this is now _incorrect_, as this is the thing that shows up on PyPI.

pyproject.toml
```diff
@@ -75,9 +75,16 @@ lab = [
...
+Source = "https://github.com/jupyterhub/jupyter-server-proxy"
+Tracker = "https://github.com/jupyterhub/jupyter-server-proxy/issues"
+
+
+# hatch ref: https://hatch.pypa.io/latest/
+#
 [tool.hatch.version]
-source = "nodejs"
-path = "labextension/package.json"
+path = "jupyter_server_proxy/_version.py"
```
bollwyvl, line 87: if using `importlib.metdata`, would be able to drop this, as the `pyproject.toml` would be the single-source-of-truth for the python distribution.

labextension/package.json
consideRatio: > npm and its raft of things has a different schema for valid versions, which would also need to be take into account.

This is valid semver, where `-0.dev` suffix will make it "earlier" than a `-beta.1` suffix. When publishing the helm charts dev releases we had to use valid semver versions and concluded this was a good dev release first part of a version suffix to avoid issues.

I'll capture the readme action point in a summary below and mark this as reslved for now.

consideRatio: Hmmm wait so the README.md file is still around and it takes precidence over the description so things would look good still I think, see https://www.npmjs.com/package/@jupyterhub/jupyter-server-proxy.

I'll add misc things back and avoid the complexity of figuring anything out in detail.

pyproject.toml
consideRatio: This was incorrect before as well I think, but I'll update it by defining a description for the PyPI package.

pyproject.toml
```diff
@@ -17,13 +16,14 @@ requires = [
...
-    "urls",
-    "version",
+description = "A JupyterLab extension accompanying the PyPI package jupyter-server-proxy adding launcher items for configured server processes."
+keywords = ["jupyter", "jupyterlab", "jupyterlab-extension"]
+authors = [
+  { name = "Ryan Lovett", email = "rylo@berkeley.edu" },
+  { name = "Yuvi Panda", email = "yuvipanda@gmail.com" },
+  { name = "Jupyter Development Team", email = "jupyter@googlegroups.com" },
 ]
+dynamic = ["version"]
```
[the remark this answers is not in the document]
consideRatio: Dynamic removed

consideRatio:
We have test errors now, but I don't think I broke it in this PR. This PR is on the other hand fixing our test suite in other regards so I suggest it gets merged to enable easier focus on resolving the other test failure about "socket already in use"

manics (approved):
Seems OK to me, but I'll let @bollwyvl make the final decision

jupyter_server_proxy/__init__.py
```diff
@@ -1,5 +1,6 @@
 from jupyter_server.utils import url_path_join as ujoin
 
+from ._version import __version__  # noqa
```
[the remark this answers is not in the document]
bollwyvl: I guess once something _has_ been hoisted to `__init__.py`, for whatever reason, it is part of the public API. Doing in-the-wild searching sounds more tiresome.

This might be an appropriate time to consider getting on the recent push to get more of the stack working with `mypy`, with a backstop of the `unused-imports` rules from `ruff`, which can more generally be informed by [sp-repo-review](https://pypi.org/project/sp-repo-review/). This is likely better than making on-the-fly, case-by-case calls on specific repos.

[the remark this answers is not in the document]
consideRatio: That sounds interesting @bollwyvl! I'm not able to drive that effort as part of this PR though, feeling somewhat overloaded atm :/

consideRatio:
@bollwyvl I'll go for a merge at this point, I hope that is considered acceptable. I'd like to see us get the projects CI system back and functional and this is just one step along the way it seems.

PR #434, opened 2023-11-17
tests: fix failure to log and raise error when waiting for url fails

consideRatio:
Trying to understand more about #435


consideRatio:
Things seem to work now, closing.

PR #437, opened 2023-12-07
Fix double formatting of callables (`command`, `environment`, `request_headers_override`)

diocas:
If a server has a callable as the environment, jupyter-server-proxy formatted it twice. 

If that callable returned an escaped var (i.e. `{ "MYVAR": "{{var}}" }`), the server was crashing when doing the second formatting (as "var" does not exist).

```python
>>> myVar = "{{var}}"
>>> myVar = myVar.format({})
>>> myVar = myVar.format({})
  File "<stdin>", line 1, in <module>
KeyError: 'var'
```

With this PR I only format once, wether it's callable or dict.

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -654,9 +654,7 @@ class NamedLocalProxyHandler(LocalProxyHandler):
     def _realize_rendered_template(self, attribute):
         """Call any callables, then render any templated values."""
         if callable(attribute):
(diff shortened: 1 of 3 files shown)


consideRatio:
I saw that this was a bug for `command`, `environment`, and `request_headers_override`, so I updated the PR title to reflect that.

Thank you @diocas for the thorough work fixing this with a description and a test!! :heart: :tada: :sunflower:

Issue #75, opened 2019-01-03
support tcp over websocket


Cas-pian:
Is there a plan to support TCP over websocket? Thanks

ryanlovett:
I've no plans to implement this directly in the extension, but @yuvipanda might.

You might look into [websockify](novnc) however. I once wrote an alpha-quality extension, [nbnovnc](https://github.com/ryanlovett/nbnovnc), which used it to proxy VNC. I need to update it to work with this extension's 1.0dev.

yuvipanda:
Thanks for opening this issue, @Cas-pian. I agree with @ryanlovett that websockify is the way to go here, and we shouldn't add that functionality into jupyter-server-proxy itself. 

Better websockify link: https://github.com/novnc/websockify

I'm going to close this one now. If you end up using websockify with a TCP server, let us know and we'll happily link to it!

Cas-pian:
Thanks for all your reply. It seems that if I want to tunnelingTCP over jupyter server's connection with I need to develop it base on tornado's websocket handler, right? thanks again.

yuvipanda:
@Cas-pian depending on what exactly you are trying to do, websockify might be enough.

Since this sounds like a wider discussion than just about this repository, can you open a post in discourse.jupyter.org? Thanks.

Cas-pian:
@yuvipanda Thanks! I'll try to take a dig on it and then try to feed back, thanks!

yuvipanda:
Re-opening per https://github.com/jupyterhub/jupyter-server-proxy/pull/447#issuecomment-2085761198

PR #447, opened 2024-02-18
Add `raw_socket_proxy` to directly proxy websockets to TCP/unix sockets

dylex:
This incorporates the functionality of websockify by allowing websockets to be proxied to simple TCP or unix socket streams. 

While I understand that websockify already provides a working solution to this problem (see discussion on #75), this seems to have a number of benefits, mainly that it eliminates the extra proxy layer and process that websockify imposes, improving latency and performance overall. It does so with a fairly small amount of code, much smaller and ismpler than websockify itself.

To simplify the instantiation of proxy handlers, this also generalizes and combines the `_make_namedproxy_handler` and `_make_supervisedproxy_handler` functions into a single `_make_proxy_handler` that dynamically determines the class to use, improving flexibility and reducing code duplication, at the cost of a little complexity.

### DIFF
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -153,6 +153,18 @@ One of:
 
 (server-process:callable-arguments)=
 
+### `websockify`
+
+_True_ to proxy only websocket connections into raw stream connections.
+_False_ (default) if the proxied server speaks full HTTP.
+
+If _True_, the proxied server is treated a raw TCP (or unix socket) server that
+does not use HTTP.
+In this mode, only websockets are handled, and messages are sent to the backend
+server as stream data. This is equivalent to running a
+[websockify](https://github.com/novnc/websockify) wrapper.
+All other HTTP requests return 405.
+
 #### Callable arguments
 
 Any time you specify a callable in the config, it can ask for any arguments it needs
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -169,7 +169,7 @@ Defaults to _True_.
 
 (server-process:callable-arguments)=
 
-### `websockify`
+### `raw_socket_proxy`
 
 _True_ to proxy only websocket connections into raw stream connections.
 _False_ (default) if the proxied server speaks full HTTP.
@@ -177,7 +177,7 @@ _False_ (default) if the proxied server speaks full HTTP.
 If _True_, the proxied server is treated a raw TCP (or unix socket) server that
 does not use HTTP.
 In this mode, only websockets are handled, and messages are sent to the backend
-server as stream data. This is equivalent to running a
+server as raw stream data. This is similar to running a
 [websockify](https://github.com/novnc/websockify) wrapper.
 All other HTTP requests return 405.
 
--- a/jupyter_server_proxy/websockify.py
+++ b/jupyter_server_proxy/websockify.py
@@ -1,0 +1,88 @@
+"""
+A simple translation layer between tornado websockets and asyncio stream
+connections.
+
+This subsumes the functionality of websockify
+(https://github.com/novnc/websockify) without needing an extra proxy hop
+or process through with all messages pass for translation.
+"""
+
+import asyncio
+
+from .handlers import NamedLocalProxyHandler, SuperviseAndProxyHandler
+
+class WebsockifyProtocol(asyncio.Protocol):
+    """
+    A protocol handler for the proxied stream connection.
+    Sends any received blocks directly as websocket messages.
+    """
+    def __init__(self, handler):
+        self.handler = handler
+
+    def data_received(self, data):
+        "Send the buffer as a websocket message."
+        self.handler._record_activity()
+        self.handler.write_message(data, binary=True) # async, no wait
+
+    def connection_lost(self, exc):
+        "Close the websocket connection."
+        self.handler.log.info(f"Websockify {self.handler.name} connection lost: {exc}")
+        self.handler.close()
+
+class WebsockifyHandler(NamedLocalProxyHandler):
+    """
+    HTTP handler that proxies websocket connections into a backend stream.
+    All other HTTP requests return 405.
+    """
+    def _create_ws_connection(self, proto: asyncio.BaseProtocol):
+        "Create the appropriate backend asyncio connection"
+        loop = asyncio.get_running_loop()
+        if self.unix_socket is not None:
+            self.log.info(f"Websockify {self.name} connecting to {self.unix_socket}")
+            return loop.create_unix_connection(proto, self.unix_socket)
+        else:
+            self.log.info(f"Websockify {self.name} connecting to port {self.port}")
+            return loop.create_connection(proto, 'localhost', self.port)
+
+    async def proxy(self, port, path):
+        raise web.HTTPError(405, "websockets only")
+
+    async def proxy_open(self, host, port, proxied_path=""):
+        """
+        Open the backend connection. host and port are ignored (as they are in
+        the parent for unix sockets) since they are always passed known values.
+        """
+        transp, proto = await self._create_ws_connection(lambda: WebsockifyProtocol(self))
+        self.ws_transp = transp
+        self.ws_proto = proto
+        self._record_activity()
+        self.log.info(f"Websockify {self.name} connected")
+
+    def on_message(self, message):
+        "Send websocket messages as stream writes, encoding if necessary."
+        self._record_activity()
+        if hasattr(self, "ws_transp"):
+            if isinstance(message, str):
+                message = message.encode('utf-8')
+            self.ws_transp.write(message) # buffered non-blocking. should block?
+
+    def on_ping(self, message):
+        "No-op"
+        self._record_activity()
+
+    def on_close(self):
+        "Close the backend connection."
+        self.log.info(f"Websockify {self.name} connection closed")
+        if hasattr(self, "ws_transp"):
+            self.ws_transp.close()
+
+class SuperviseAndWebsockifyHandler(SuperviseAndProxyHandler, WebsockifyHandler):
+    async def _http_ready_func(self, p):
+        # not really HTTP here, just try an empty connection
+        try:
+            transp, _ = await self._create_ws_connection(asyncio.Protocol)
+        except OSError as exc:
+            self.log.debug(f"Websockify {self.name} connection check failed: {exc}")
+            return False
+        transp.close()
+        return True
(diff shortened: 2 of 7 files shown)


manics:
Thanks for the PR, and sorry for the delay in reviewing. I like the idea of having a self contained pure Python implementation! I can't guarantee we'll merge it since `websockify` works, seems reliable, and it's always nice if someone else maintains things, but https://github.com/jupyterhub/jupyter-server-proxy/issues/75 is several years old and I think we should consider this.

To help with review would you mind rebasing this onto `main`? We [recently fixed a security vulnerability](https://github.com/jupyterhub/jupyter-server-proxy/blob/main/docs/source/changelog.md), so it would be helpful when testing locally if we could just checkout an up-to-date branch with your changes. Thanks!

dylex:
Thanks, yeah I understand the concerns, but having worked on the websockify code as well, I can at least say that this PR is overall much simpler and easier to maintain, and I'm happy to help do so.

I've rebased, and tested again on the updated version (with jupyter-remote-desktop-proxy) and things look okay. Probably there should be some automated tests for this, too -- I can look into that if there's interest in moving forward.

manics:
Thanks! Tests would be great if we go ahead. I'll try out your changes locally first, feel free to bump this issue if I forget!

yuvipanda:
Bump reminder, @manics :)

manics (commented):
Sorry for the delay in reviewing this. The code looks OK to me, but I don't have much experience with websockets so ideally it could do with a second review.

I've tested this with jupyter-remote-desktop-proxy and it works well!
```diff
diff --git a/jupyter_remote_desktop_proxy/setup_websockify.py b/jupyter_remote_desktop_proxy/setup_websockify.py
index c6b1f4b..f25801e 100644
--- a/jupyter_remote_desktop_proxy/setup_websockify.py
+++ b/jupyter_remote_desktop_proxy/setup_websockify.py
@@ -12,8 +12,6 @@ def setup_websockify():
         raise RuntimeError(
             "vncserver executable not found, please install a VNC server"
         )
-    if not which('websockify'):
-        raise RuntimeError("websockify executable not found, please install websockify")
 
     # TurboVNC and TigerVNC share the same origin and both use a Perl script
     # as the executable vncserver. We can determine if vncserver is TigerVNC
@@ -30,16 +28,10 @@ def setup_websockify():
         is_tigervnc = "tigervnc" in vncserver_file.read().casefold()
 
     if is_tigervnc:
-        # Make a secure temporary directory for sockets that is only readable,
-        # writeable, and searchable by our uid - TigerVNC can listen to a Unix
-        # socket!
-        sockets_dir = tempfile.mkdtemp()
-        sockets_path = os.path.join(sockets_dir, 'vnc-socket')
-
-        websockify_args = ['--unix-target', sockets_path]
-        vnc_args = [vncserver, '-rfbunixpath', sockets_path]
+        unix_socket = True
+        vnc_args = [vncserver, '-rfbunixpath', "{unix_socket}"]
     else:
-        websockify_args = []
+        unix_socket = False
         vnc_args = [vncserver, '-localhost', '-rfbport', '{port}']
 
     if not os.path.exists(os.path.expanduser('~/.vnc/xstartup')):
@@ -58,18 +50,13 @@ def setup_websockify():
     )
 
     return {
-        'command': [
-            'websockify',
-            '--verbose',
-            '--heartbeat=30',
-            '{port}',
-        ]
-        + websockify_args
-        + ['--', '/bin/sh', '-c', f'cd {os.getcwd()} && {vnc_command}'],
+        'command': ['/bin/sh', '-c', f'cd {os.getcwd()} && {vnc_command}'],
         'timeout': 30,
         'new_browser_window': True,
         # We want the launcher entry to point to /desktop/, not to /desktop-websockify/
         # /desktop/ is the user facing URL, while /desktop-websockify/ now *only* serves
         # websockets.
         "launcher_entry": {"title": "Desktop", "path_info": "desktop"},
+        "unix_socket": unix_socket,
+        "websockify": True,
     }
```

@yuvipanda what do you think?

image

yuvipanda:
I very much like the idea of not depending on websockify for TCP streams. This would allow us to package remote-desktop-proxy in conda-forge more easily too. I've re-opened #75 and will look through the code now.

jupyter_server_proxy/websockify.py
```diff
@@ -0,0 +1,88 @@
...
+    A protocol handler for the proxied stream connection.
+    Sends any received blocks directly as websocket messages.
+    """
+    def __init__(self, handler):
+        self.handler = handler
+
+    def data_received(self, data):
+        "Send the buffer as a websocket message."
+        self.handler._record_activity()
+        self.handler.write_message(data, binary=True) # async, no wait
```
yuvipanda, line 25: Looking at https://www.tornadoweb.org/en/stable/websocket.html#tornado.websocket.WebSocketHandler.write_message, I see that it does return a `Future`. Can you help me understand why this doesn't need to be awaited here? I see that `data_received` is not a coroutine, and Guido explains [why](https://groups.google.com/g/python-tulip/c/z-IVH5RoDzo/m/SpZc0zTuPJsJ). But I don't fully understand what happens here if `data_received` is called twice with data (in order), but the previous `write_message` has not completed (because `write_message` is async). Would this not garble the data and have it be transmited out of order? Glyph's comment further down that thread has the same question as me...

If you think this isn't going to cause ordering problems, adding a comment explaining why would be very helpful!

docs/source/server-process.md
```diff
@@ -153,6 +153,18 @@ One of:
 
 (server-process:callable-arguments)=
 
+### `websockify`
```
yuvipanda, line 156: Let's call this something else - perhaps `raw_tcp_proxy_only` (or something else, am not tied to that)? Otherwise there'll be confusion about what it means to 'use jupyter-server-proxy with websockify'.

yuvipanda (changes requested):
I'll do another round next week, but left two comments here!

I also think we need to add a test - perhaps similar to the existing websocket test, but with a simple TCP server?

Am really excited to get this in, as it would allow us to do a bunch of other cool stuff elsewhere :) For example, it would enable proper ssh forwarding in https://github.com/yuvipanda/jupyterhub-ssh...

Thank you for working on this, and your patience as we review this, @dylex!

docs/source/server-process.md
yuvipanda: Oh, also mention that this would then bypass the code we have that allows folks to rewrite responses.

jupyter_server_proxy/websockify.py
dylex: This also threw me off but digging into tornado a bit I'm pretty sure it's okay. `write_message` queues up the message before doing anything async, and the async returned is meant only to be used (optionally) for control flow, so ordering is preserved, even if the first messages hasn't been sent. There's a bit of commentary on this [here](https://github.com/tornadoweb/tornado/blob/bdfc017c66817359158185561cee7878680cd841/tornado/websocket.py#L1073). There may be an issue determining exactly which messages have been acknowledged if the connection is closed or there's another error, but given that that will abort the whole thing, and there's no way to communicate that anyway, that should be fine. I'll try to clarify this in comments.

docs/source/server-process.md
dylex: Happy to rename it whatever people feel makes sense.

It doesn't exactly bypass the response rewriting, but the rewriting only is used for normal http responses anyway, not websockets, so because all other requests return 405 it never gets used. I guess we could call rewrite on the 405 response? But yes, I'll clarify this, too.

jupyter_server_proxy/websockify.py
yuvipanda: https://github.com/tornadoweb/tornado/blob/bdfc017c66817359158185561cee7878680cd841/tornado/websocket.py#L1073 is perfect, include it in the comment as well and I think we're good.

docs/source/server-process.md
yuvipanda: I don't want to get into too much of a naming rabbit hole, but I'll suggest `raw_tcp_proxy` here (without the `only`). Any objections to that, @dylex?

jupyter_server_proxy/websockify.py
```diff
@@ -0,0 +1,89 @@
...
+        loop = asyncio.get_running_loop()
+        if self.unix_socket is not None:
+            self.log.info(f"Websockify {self.name} connecting to {self.unix_socket}")
+            return loop.create_unix_connection(proto, self.unix_socket)
+        else:
+            self.log.info(f"Websockify {self.name} connecting to port {self.port}")
+            return loop.create_connection(proto, 'localhost', self.port)
+
+    async def proxy(self, port, path):
+        raise web.HTTPError(405, "websockets only")
```
yuvipanda, line 49: Let's raise a more descriptive error message here, including the fact that `raw_tcp_proxy` was specified?

jupyter_server_proxy/websockify.py
```diff
@@ -0,0 +1,89 @@
...
+        transp, proto = await self._create_ws_connection(lambda: WebsockifyProtocol(self))
+        self.ws_transp = transp
+        self.ws_proto = proto
+        self._record_activity()
+        self.log.info(f"Websockify {self.name} connected")
+
+    def on_message(self, message):
+        "Send websocket messages as stream writes, encoding if necessary."
+        self._record_activity()
+        if hasattr(self, "ws_transp"):
```
yuvipanda, line 65: Is there a reason to add this guard here explicitly, given that `on_message` would not be called before `proxy_open`?

yuvipanda (changes requested):
I spent a little more time looking at this, and the overall structure seems fine to me. It does need a test (perhaps modeled after what our current websocket tests do, but with raw TCP). After that + the naming stuff, happy to merge.

Thank you for working through this, @dylex!

docs/source/server-process.md
dylex: Sounds fine to me. It is specifically using websockets but I guess that should be clear.

jupyter_server_proxy/websockify.py
dylex: I think that's true as it stands, though on reviewing I realize `proxy_open` now should probably do a `_check_host_allowlist` and maybe other checks that may result in different situations. I'll look at it again with the latest code and clean up if possible.

dylex: Or wait, no that's not relevant here since we're always proxying to a fixed place. (Sorry, haven't looked at this for a bit.)

docs/source/server-process.md
yuvipanda: @dylex Let's do `raw_socket_proxy`!

dylex:
Okay, I think everything has been addressed now.  Option is now named `raw_socket_proxy`, and there are a couple tests with a simple tcp and unix socket server backend. Let me know if there's anything I should cleanup/rebase.

yuvipanda:
Alright, thank you very much for your work, @dylex! And sorry about the slow review. I hope to see more contributions from you in the future too <3

Issue #449, opened 2024-02-19
Test failures in main branch


consideRatio:
There are acceptance test failures in main branch (example [here](https://github.com/jupyterhub/jupyter-server-proxy/actions/runs/7957862509/job/21721593149)).

A key part of the failure is:

```
==============================================================================
Acceptance.Notebook :: Server Proxies in Notebook                             
==============================================================================
Notebook Loads                                                        | PASS |
------------------------------------------------------------------------------
Launch Browser Tab                                                    | FAIL |
Element with locator 'xpath://div[contains(@class, "jp-FileBrowser-toolbar")]//*[contains(text(), "New")]' not found.
------------------------------------------------------------------------------
Launch Another Browser Tab                                            | FAIL |
Element with locator 'xpath://div[contains(@class, "jp-FileBrowser-toolbar")]//*[contains(text(), "New")]' not found.
```

They seem to show up due to this part:

https://github.com/jupyterhub/jupyter-server-proxy/blob/9b624c4d9507176334b46a85d94a4aa3bcd29bed/tests/acceptance/Notebook.robot#L11-L13

And specifically with `notebook==7.1.1` but not with `notebook==7.0.8`.

I understand those expressions to be selecting a specific HTML element, but I think if it worked once it should work later as I see no difference in the HTML looking at it with old/new versions.

![image](https://github.com/jupyterhub/jupyter-server-proxy/assets/3837114/805234f1-df33-4840-b2b5-da123329556d)
![image](https://github.com/jupyterhub/jupyter-server-proxy/assets/3837114/0953db03-023d-406f-be50-7edacd2aad05)

consideRatio:
@bollwyvl I'm stumped, it sais `Element with locator 'xpath://div[contains(@class, "jp-FileBrowser-toolbar")]//*[contains(text(), "New")]' not found.` when using `notebook==7.1.*` but not when using `notebook==7.0.*`.

I tried opening a chrome inspector console and doing `$x(......)` stuff in order to see if I could reproduce the difference, but I ended up getting an empty `[]` response no matter what. I don't understand xpath expressions yet or if this debugging strategy was reasonable =/

Do you have ideas on debugging / resolving this issue?

bollwyvl:
yes, 4.1 and 7.1 created breaking changes in the DOM.

consideRatio:
@bollwyvl ah thanks, I'm making progress now!

manics:
[...]

PR #452, opened 2024-02-20
ci: schedule test runs twice a week

consideRatio:
We had test failures in main branch (#449), and realizing that something has started failing is often something that goes with tracking what dependencies has changed.

If I think there is a test that has started failing for some reason outside a code change in repo, I do the following:

1. Look at workflows run recently to identify when they last worked and when they started failing
2. Compare `pip freeze` output between succeeding tests and failing tests by copy pasting into `ok.txt` and `fail.txt` and doing `git diff --no-index -U0 -- ok.txt fail.txt` (-U0 means 0 lines of context surrounding changes observed)

This is troublesome if there isn't a workflow run recently though, but having regular runs, for example twice a week, would do the trick.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,6 +21,9 @@ on:
     branches-ignore:
       - "dependabot/**"
       - "pre-commit-ci-update-config"
+  schedule:
+    # Run at 05:00 on monday and thursday, ref: https://crontab.guru/#0_5_*_*_1,4
+    - cron: "0 5 * * 1,4"
   workflow_dispatch:
 
 env:


consideRatio:
In https://github.com/jupyterhub/jupyter-server-proxy/issues/449#issuecomment-1954714794 @manics wrote:

> I think this could be a good case for using dependabot to pin some dependencies for CI testing. Keep the dev dependencies unrestricted in pyproject.toml, but use a separate dev-ci-requirements.txt file for CI only and let dependabot tell us when there's a UI change in lab or notebook.

@manics I prefer settling for doing this (ensure more regular runs of tests) over adding dependabot to work on a CI specific dependency file as it avoids some complexity and maintenance chore. What do you think?

PR #244, opened 2021-02-08
Add configurable choice of a JupyterLab launcher category to present registered proxies (Notebook, Console, Other)

eddienko:
As preparation work for #145 this pull request adds the capability to configure under which launcher category the proxy applications will appear. 

It needs a bit of check from people who have more experience with extensions.
Screenshots follow.

Screenshot 2021-02-08 at 20 54 39

Screenshot 2021-02-08 at 20 55 23

Screenshot 2021-02-08 at 20 55 53

PR #453, opened 2024-02-20
Allow configuring jupyterlab launcher category

dylex:
This lets you set which row of the launcher the icon ends up on (Notebook, Console, or Other).  Just pass a configured launcher_entry.category through to launcher.add, defaulting to Notebook. Also adds documentation for existing `path_info` which was missing.

### DIFF
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -123,9 +123,15 @@ the following keys:
    explicit entry.
 2. **icon_path**
    Full path to an svg icon that could be used with a launcher. Currently only used by the
-   JupyterLab launcher
+   JupyterLab launcher, when category is "Notebook" (default) or "Console".
 3. **title**
    Title to be used for the launcher entry. Defaults to the name of the server if missing.
+4. **path_info**
+   The trailing path that is appended to the user's server URL to access the proxied server.
+   By default it is the name of the server followed by a trailing slash.
+5. **category**
+   The category for the launcher item. Currently only used by the JupyterLab launcher.
+   By default it is "Notebook".
 
 ### `new_browser_tab`
 
--- a/jupyter_server_proxy/api.py
+++ b/jupyter_server_proxy/api.py
@@ -22,6 +22,7 @@ class ServersInfoHandler(JupyterHandler):
                     "enabled": sp.launcher_entry.enabled,
                     "title": sp.launcher_entry.title,
                     "path_info": sp.launcher_entry.path_info,
+                    "category": sp.launcher_entry.category,
                 },
                 "new_browser_tab": sp.new_browser_tab,
             }
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -24,7 +24,7 @@ except ImportError:
 
 
 LauncherEntry = namedtuple(
-    "LauncherEntry", ["enabled", "icon_path", "title", "path_info"]
+    "LauncherEntry", ["enabled", "icon_path", "title", "path_info", "category"]
 )
 ServerProcess = namedtuple(
     "ServerProcess",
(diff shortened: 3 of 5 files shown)


dylex:
I guess I should add (and probably document) that for [obscure jupyterlab reasons](https://github.com/jupyterlab/jupyterlab/blob/cfa3f5b94ecdddef8a085b397963e87a0066b559/packages/launcher/src/widget.tsx#L119), the `icon_path` setting will only work when `category` is either `"Kernel"` or `"Console"`.

docs/source/server-process.md
```diff
@@ -126,6 +126,12 @@ the following keys:
    JupyterLab launcher
 3. **title**
    Title to be used for the launcher entry. Defaults to the name of the server if missing.
+4. **path_info**
+   The trailing path that is appended to the user's server URL to access the proxied server.
+   By default it is the name of the server followed by a trailing slash.
+5. **category**
+   The category for the launcher item. Currently only used by the JupyterLab launcher.
+   By default it is "Notebook".
```
yuvipanda, line 134: I think if we can indicate here that icons won't work unless category is 'Notebook' or 'Kernel', we can get this merged as is!

bollwyvl:
> the icon_path setting will only work

Non-kernel icons in more recent jupyter clients are SVG so that they can be themed e.g. a black icon on a dark theme would disappear.

It's certainly possible to generate new SVG icons that answer the contract on the fly. Even PNG/JPEG/etc. can be encapsulated when the [command is registered](https://github.com/jupyterhub/jupyter-server-proxy/blob/ca21c38c9dee85703df94cf4f3222e0ce04fc047/labextension/src/index.ts#L164); very naively:

```ts
commands.addCommand(CommandIDs.open, {
  label: (args) => (args as IOpenArgs).title,
  icon: new LabIcon({
    svgstr: `<svg><foreignObject><img src="${launcher_entry.icon_url}"/></foreignObject></svg>`
  }),
  ...
})
```

...though they won't somehow magically be theme-aware. 

Some gotchas:
- it may also be that the path would need some massaging with e.g. `new URL(launcher_entry.icon_url, PageConfig.getBaseUrl())`
- the `svg` might need `viewbox`
- the `img` might need a `width` and `height`

yuvipanda (approved):
I don't currently have time to test this locally, but if someone else can I think we should merge this as is! @bollwyvl's suggestions about images perhaps can be done in a different PR? @dylex would that be something you're interested in doing?

dylex:
I looked into the other icons a little, and it seemed more complicated. I think @bollwyvl's suggestion could be made to work in a generic way.  I can try to take a look at some point, but probably not for this PR. 

I also just noticed there was an old draft PR #244 to provide this same functionality in a similar way -- hopefully not stepping on anyone's toes!

imcovangent:
Hello!

This is exactly what I was looking for! :) Plus I also wanted to add rank and make the SVG icon work. You can find that here: https://github.com/imcovangent/jupyter-server-proxy/tree/category_and_rank

I'm not sure what the proper way is to include this. I have now forked dylex:category and created a new branch that builds upon this PR, but adds the rank argument as well. And it includes a fix for the icons.

What is best? I make this a PR on dylex or make it directly a (new?) PR here?

Or do I wait for this PR to complete and then make a new PR? Please advice @dylex @yuvipanda

imcovangent:
Any feedback on this @dylex or @yuvipanda ? Else I will open a new PR here from my branch. I'd love to see my additions in the next release.

yuvipanda:
I just tested this, and it works. It has the limitation that the icon doesn't show up if you don't put it in `Notebook`, but that could be fixed in another PR.

yuvipanda:
@imcovangent I opened a PR from your fork in https://github.com/jupyterhub/jupyter-server-proxy/pull/477, will test that too

yuvipanda:
@imcovangent if we can make the SVG setup from you work, I'm happy to review, merge it and then get a new release out.

PR #448, opened 2024-02-18
Set subprotocols=None when there is no subprotocol proposed

benz0li:
- Modify tests to use headers
- Cleanup test class SubprotocolWebSocket (Edited to clarify its a test class only)
- Add test for empty and no subprotocols
- Fix https://github.com/jupyterhub/jupyter-server-proxy/issues/442
- Fix https://github.com/jupyterhub/jupyter-server-proxy/issues/445

## EDIT

- We think this fixes #332, and will assume that for now but could be wrong.

PR #457, opened 2024-02-21
List tornado and traitlets as dependencies explicitly, and cleanup unreachable code

consideRatio:
We reference tornado and traitlets directly from our source code, so we should surface those dependecies.

We require `jupyter-server>=1.0`, which [depend on `tornado>=5.0` and `traitlets>=4.2.1`](https://github.com/jupyter-server/jupyter_server/blob/1.0.0/setup.py#L39), so I'm adding our dependency on tornado and traitlets explicitly in this project, putting their lower bounds to those versions - and adding a test to verify our test suite works against those versions.

For reference, note that `jupyter-server==1.1.0` released four years ago requires tornado 6.1, so I figure its safe to assume we can bump these notably without issues in the future.

---

This enables me to better reason about https://github.com/jupyterhub/jupyter-server-proxy/pull/448, as its changing something that is passed to tornado, and tornado's behavior may differ slightly between versions.

### DIFF
--- a/jupyter_server_proxy/websocket.py
+++ b/jupyter_server_proxy/websocket.py
@@ -7,7 +7,7 @@ Some original inspiration from https://github.com/senko/tornado-proxy
 import inspect
 
 from jupyter_server.utils import ensure_async
-from tornado import httpclient, httputil, ioloop, version_info, websocket
+from tornado import httpclient, httputil, websocket
 
 
 class PingableWSClientConnection(websocket.WebSocketClientConnection):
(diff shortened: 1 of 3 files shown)


jupyter_server_proxy/websocket.py
```diff
@@ -40,31 +40,21 @@ def pingable_ws_connect(
...
 
-    # for tornado 4.5.x compatibility
-    if version_info[0] == 4:
-        conn = PingableWSClientConnection(
-            io_loop=ioloop.IOLoop.current(),
-            compression_options={},
-            request=request,
-            on_message_callback=on_message_callback,
-            on_ping_callback=on_ping_callback,
-        )
```
consideRatio, line 51: This is cleaned up now that we know we required tornado 5.0 via jupyter server 1.0.0+.

jupyter_server_proxy/websocket.py
```diff
@@ -40,31 +40,21 @@ def pingable_ws_connect(
...
+        request=request,
+        compression_options={},
+        on_message_callback=on_message_callback,
+        on_ping_callback=on_ping_callback,
+        max_message_size=getattr(
+            websocket, "_default_max_message_size", 10 * 1024 * 1024
+        ),
+        subprotocols=subprotocols,
+        **kwargs,
+    )
```
consideRatio, line 57: This was just indentation changes by no longer nesting under an `else` clause.

minrk (approved):

PR #473, opened 2024-04-15
feat: Allow Skipping of Activity Reporting

### LINKED ISSUE #471: Ability to Configure Activity Reporting
### Proposed change
Add ability to customize or disable activity reporting to the JupyterHub handler. Currently, all requests to the proxy get marked as activity. This means if you leave your browser open with an open tab that goes through the proxy and don't interact with the page your JupyterHub pod will not get culled!

Anyone who wants to leverage JupyterHub pod culling that has a better was of detecting activity.

sdmichelini:
Allows proxy to be configured if it reports back to JupyterHub. In some cases the proxy will always detect background activity which means that the pod will never get culled. In this case you must disable recording and use an alternative source of activity(outside of this MR).

Closes #471

### DIFF
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -157,6 +157,16 @@ One of:
 - A callable that takes any {ref}`callable arguments <server-process:callable-arguments>`,
   and returns a dictionary of strings that are used & treated same as above.
 
+### `skip_activity_reporting`
+
+Whether to report activity from the proxy to JupyterHub. If _True_, JupyterHub
+will not be notified of new activity.
+
+Useful if you want to have a seperate way of determining activity through a
+proxied application.
+
+Defaults to _False_.
+
 (server-process:callable-arguments)=
 
 #### Callable arguments
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -160,7 +160,7 @@ One of:
 ### `update_last_activity`
 
 Whether to report activity from the proxy to Jupyter Server. If _True_, Jupyter Server
-will not be notified of new activity.
+will be notified of new activity. This is primarily used by JupyterHub for idle detection and culling.
 
 Useful if you want to have a seperate way of determining activity through a
 proxied application.
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -157,15 +157,15 @@ One of:
 - A callable that takes any {ref}`callable arguments <server-process:callable-arguments>`,
   and returns a dictionary of strings that are used & treated same as above.
 
-### `skip_activity_reporting`
+### `update_last_activity`
 
-Whether to report activity from the proxy to JupyterHub. If _True_, JupyterHub
+Whether to report activity from the proxy to Jupyter Server. If _True_, Jupyter Server
 will not be notified of new activity.
 
 Useful if you want to have a seperate way of determining activity through a
 proxied application.
 
-Defaults to _False_.
+Defaults to _True_.
 
 (server-process:callable-arguments)=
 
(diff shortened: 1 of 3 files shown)


ryanlovett:
Thanks for the contribution @sdmichelini ! This seems like it would be useful if you want the culler to ignore a proxied app's activity, and I think your change will have the effect you intend.

Could you also please add an entry in `doc/source/server-process.md`?

Lastly, it'd also be great if you could add a test case under `tests/`.

sdmichelini:
@ryanlovett Thanks for the prompt review! I've added documentation for the new parameter in the suggested location.

As for the tests is there a config-test pattern to follow that you would recommend? I did a check over the current tests and don't see ones for config where I could plug in other than confirm basic launching of the HTTP server.

ryanlovett:
@sdmichelini I think you'd need to create a server to proxy that generates "spurious" activity, such as the kind that you want to mask via this PR. `resources/httpinfo.py` is a simple server, but for this test implementation you'd need to write one that is more active. That might be the most difficult part.

Then you'd need to add an entry to `jupyter_server_config.py` that sets the new configuration option and proxies the busy test server, and write a test function in `test_proxies.py` that queries the proxy for activity. It would need to confirm that there is no activity even when you might expect there to be some.

I believe the hub uses [/api/status](https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html#get--api-status) to check for activity. I would imagine `last_activity` is not updated frequently, or at all, when you are not proxying your real world busy app, but is updated when that app is running. If so then your test would confirm that, with your new configuration option set, `last_activity` is not updated even when the busy test server is running.

This is all easier for me to say than do, so let me know if you have any questions.

ryanlovett:
No worries. Are you able to share info about what app you were proxying? (in the event that we can simulate its behavior)

sdmichelini:
Yes - https://github.com/coder/code-server - is the app I am proxying.

sdmichelini:
@ryanlovett any objections to adding this? would be useful to have this feature.

deser:
Looking forward for this feature as well.

ryanlovett:
@sdmichelini No objections. Let me add another reviewer...

yuvipanda (changes requested):
Thank you for the PR @sdmichelini and thank you for tagging me in, @ryanlovett.

This looks good to me, but I'd like to request the name of this be changed. The activity tracking is really a part of jupyter_server, and can be used by any application (not just JupyterHub, although it's the most common user).

In jupyter_server (and in JupyterHub too I think) it's called 'last_activity'. 

So I suggest the following course of action:

1. Change the name of the property to `update_last_activity`
2. When set to `True`, it will retain current default behavior. When set to `False`, it will no longer update `last_activity`. This is an inverse of the current config setting. The default will be `True` (inversion of this PR's default again)

I'm happy to merge this once this change is done.

Thanks!

sdmichelini:
@yuvipanda Thank you for taking a look! I have updated the pull-request based on your feedback!

docs/source/server-process.md
```diff
@@ -157,6 +157,16 @@ One of:
 - A callable that takes any {ref}`callable arguments <server-process:callable-arguments>`,
   and returns a dictionary of strings that are used & treated same as above.
 
+### `update_last_activity`
+
+Whether to report activity from the proxy to Jupyter Server. If _True_, Jupyter Server
+will not be notified of new activity.
```
yuvipanda, line 163:
```suggestion
will be notified of new activity. This is primarily used by JupyterHub for idle detection and culling
```

yuvipanda:
Thanks a lot, @sdmichelini! Looking forward to more PRs from you :)

PR #477, opened 2024-06-12
Added rank and fixed SVG icon handling

yuvipanda:


imcovangent:
Hi @yuvipanda,

I also found out it is quite sensitive. It only works for SVG files and even then only if the SVG files really only start with the <svg> element and hence end with </svg>.

I will have a look into this later to make it more versatile.

Will ping you if I have a PR ready for that. Will be in a couple of weeks probably though.

PR #478, opened 2024-06-15
Fixed jlpm to jupyter in bash command

imcovangent:
Fixed command to link your development version of the extension with JupyterLab and Notebook.


yuvipanda:
Thank you, @imcovangent!

PR #479, opened 2024-06-18
feat: support stream api

ganisback:
currently jupyter-server-proxy does not support the text/event-stream api. when we install https://github.com/hiyouga/LLaMA-Factory in the notebook instance, and access by : http://xxx.xxx/proxy/7860/
the normal request works fine, but the stream API never response.
reference code: https://github.com/ideonate/jhsingle-native-proxy/blob/3e26a4ee0e7318970b7cf6abbd7d88455a9ac621/jhsingle_native_proxy/proxyhandlers.py#L217

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -4,7 +4,7 @@ Authenticated HTTP proxy for Jupyter Notebooks
 Some original inspiration from https://github.com/senko/tornado-proxy
 """
 
-import os
+import os, json, re
 import socket
 from asyncio import Lock
 from copy import copy
@@ -287,7 +287,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
 
         return client_uri
 
-    def _build_proxy_request(self, host, port, proxied_path, body):
+    def _build_proxy_request(self, host, port, proxied_path, body, **extra_opts):
         headers = self.proxy_request_headers()
 
         client_uri = self.get_client_uri("http", host, port, proxied_path)
@@ -307,6 +307,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
             decompress_response=False,
             headers=headers,
             **self.proxy_request_options(),
+             **extra_opts,
         )
         return req
 
@@ -365,7 +366,82 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
                 body = b""
             else:
                 body = None
+        accept_type = self.request.headers.get('Accept')
+        if accept_type == 'text/event-stream':
+            return await self._proxy_progressive(host, port, proxied_path, body)
+        else:
+            return await self._proxy_normal(host, port, proxied_path, body)
+    
+    async def _proxy_progressive(self, host, port, proxied_path, body):
+        # Proxy in progressive flush mode, whenever chunks are received. Potentially slower but get results quicker for voila
 
+        client = httpclient.AsyncHTTPClient()
+
+        # Set up handlers so we can progressively flush result
+
+        headers_raw = []
+
+        def dump_headers(headers_raw):
+            for line in headers_raw:
+                r = re.match('^([a-zA-Z0-9\-_]+)\s*\:\s*([^\r\n]+)[\r\n]*$', line)
+                if r:
+                    k,v = r.groups([1,2])
+                    if k not in ('Content-Length', 'Transfer-Encoding',
+                                  'Content-Encoding', 'Connection'):
+                        # some header appear multiple times, eg 'Set-Cookie'
+                        self.set_header(k,v)
+                else:
+                    r = re.match('^HTTP[^\s]* ([0-9]+)', line)
+                    if r:
+                        status_code = r.group(1)
+                        self.set_status(int(status_code))
+            headers_raw.clear()
+
+        # clear tornado default header
+        self._headers = httputil.HTTPHeaders()
+
+        def header_callback(line):
+            headers_raw.append(line)
+
+        def streaming_callback(chunk):
+            # Do this here, not in header_callback so we can be sure headers are out of the way first
+            dump_headers(headers_raw) # array will be empty if this was already called before
+            self.write(chunk)
+            self.flush()
+
+        # Now make the request
+
+        req = self._build_proxy_request(host, port, proxied_path, body, 
+                    streaming_callback=streaming_callback, 
+                    header_callback=header_callback)
+
+        try:
+            response = await client.fetch(req, raise_error=False)
+        except httpclient.HTTPError as err:
+            if err.code == 599:
+                self._record_activity()
+                self.set_status(599)
+                self.write(str(err))
+                return
+            else:
+                raise
+
+        # record activity at start and end of requests
+        self._record_activity()
+
+        # For all non http errors...
+        if response.error and type(response.error) is not httpclient.HTTPError:
+            self.set_status(500)
+            self.write(str(response.error))
+        else:
+            self.set_status(response.code, response.reason) # Should already have been set
+
+            dump_headers(headers_raw) # Should already have been emptied
+
+            if response.body: # Likewise, should already be chunked out and flushed
+                self.write(response.body)
+
+    async def _proxy_normal(self, host, port, proxied_path, body):
         if self.unix_socket is not None:
             # Port points to a Unix domain socket
             self.log.debug("Making client for Unix socket %r", self.unix_socket)
@@ -457,6 +533,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
 
             if rewritten_response.body:
                 self.write(rewritten_response.body)
+
 
     async def proxy_open(self, host, port, proxied_path=""):
         """
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -374,7 +374,7 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
                 force_instance=True, resolver=UnixResolver(self.unix_socket)
             )
         else:
-            client = httpclient.AsyncHTTPClient()
+            client = httpclient.AsyncHTTPClient(force_instance=True)
         # check if the request is stream request
         accept_header = self.request.headers.get('Accept')
         if accept_header == 'text/event-stream':
@@ -425,7 +425,8 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
                     header_callback=header_callback)
         
         # no timeout for stream api
-        req.request_timeout = 0
+        req.request_timeout = 7200
+        req.connect_timeout = 600
         
         try:
             response = await client.fetch(req, raise_error=False)
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -376,8 +376,8 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
         else:
             client = httpclient.AsyncHTTPClient()
         # check if the request is stream request
-        proxy_streaming = self.request.headers.get('Accept')
-        if proxy_streaming == 'text/event-stream':
+        accept_header = self.request.headers.get('Accept')
+        if accept_header == 'text/event-stream':
             return await self._proxy_progressive(host, port, proxied_path, body, client)
         else:
             return await self._proxy_buffered(host, port, proxied_path, body, client)
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -414,7 +414,10 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
         req = self._build_proxy_request(host, port, proxied_path, body, 
                     streaming_callback=streaming_callback, 
                     header_callback=header_callback)
-
+        
+        # no timeout for stream api
+        req.request_timeout = 0
+        
         try:
             response = await client.fetch(req, raise_error=False)
         except httpclient.HTTPError as err:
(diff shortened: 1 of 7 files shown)


ganisback:
@yuvipanda please help review this feature

manics:
@ganisback Thanks for your contribution! To help with review please could you expand the PR description to explain what this does, and why it's needed? We'll need some tests if this is going to be merged, but those can be added later if we agree this feature should be added.

ganisback:
@manics  description updated

yuvipanda:
Ah interesting! I think @jmunroe also ran into this recently.

If my understanding is right, the existing code doesn't stream but waits for the entire upstream request to be completed before passing it on. And with this PR, we're special casing the `text/eventstream` to be streaming.

Instead, what do you think about modifying the existing code (what you have as `_proxy_normal`) to be streaming? I don't think there's any reason to not stream *all* requests - I think we initially waited for requests to complete simply because that was the easier thing to do. That would also help with tests - the current tests should catch that and if they pass, we can probably merge that.

So I propose that instead of only streaming eventstreams, we stream everything.

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
...
+
+        client = httpclient.AsyncHTTPClient()
+
+        # Set up handlers so we can progressively flush result
+
+        headers_raw = []
+
+        def dump_headers(headers_raw):
+            for line in headers_raw:
+                r = re.match('^([a-zA-Z0-9\-_]+)\s*\:\s*([^\r\n]+)[\r\n]*$', line)
```
yuvipanda, line 384: Can you use https://www.tornadoweb.org/en/stable/httputil.html#tornado.httputil.HTTPHeaders.parse_line here, and if that fails assume it's a status code? regexes in places like this always worry me :D

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
...
+            if err.code == 599:
+                self._record_activity()
+                self.set_status(599)
+                self.write(str(err))
+                return
+            else:
+                raise
 
+        # record activity at start and end of requests
+        self._record_activity()
```
yuvipanda, line 431: I think this should instead happen with every `self.flush` (in line 408)

yuvipanda:
As for testing, I suggest:

1. Create a simple eventstream emitting server in https://github.com/jupyterhub/jupyter-server-proxy/tree/main/tests/resources (similar to the websocket one perhaps)
2. Create tests in `test_proxy` similar to the websocket tests (https://github.com/jupyterhub/jupyter-server-proxy/blob/234a192793a6a46bc8df18f730ea0060e6d0c3e8/tests/test_proxies.py#L346).

This can be simpler than websockets, because eventstreams are simpler.

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
                 body = b""
             else:
                 body = None
+        accept_type = self.request.headers.get('Accept')
+        if accept_type == 'text/event-stream':
+            return await self._proxy_progressive(host, port, proxied_path, body)
```
yuvipanda, line 369: I suggest naming this `proxy_streaming`

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
...
+        else:
+            return await self._proxy_normal(host, port, proxied_path, body)
```
yuvipanda, line 371: I suggest naming this `proxy_buffered`

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
                 body = b""
             else:
                 body = None
+        accept_type = self.request.headers.get('Accept')
+        if accept_type == 'text/event-stream':
```
yuvipanda, line 368: Add a comment here on *why* this is needed

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
...
+    async def _proxy_progressive(self, host, port, proxied_path, body):
+        # Proxy in progressive flush mode, whenever chunks are received. Potentially slower but get results quicker for voila
+
+        client = httpclient.AsyncHTTPClient()
+
+        # Set up handlers so we can progressively flush result
+
+        headers_raw = []
+
+        def dump_headers(headers_raw):
```
yuvipanda, line 382: let's move this function outside `_proxy_progressive`. `headers_raw` as used inside this function is an arg, while there's also a local variable with the same name. Moving it outside cleans this up a little.

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
...
+            for line in headers_raw:
+                r = re.match('^([a-zA-Z0-9\-_]+)\s*\:\s*([^\r\n]+)[\r\n]*$', line)
+                if r:
+                    k,v = r.groups([1,2])
+                    if k not in ('Content-Length', 'Transfer-Encoding',
```
yuvipanda, line 387: This is a different list than the one in line 531. Is there a specific reason? If not, I think we should move this to a constant at the top of the file and reference it from both places so it doesn't get out of date.

jupyter_server_proxy/handlers.py
```diff
@@ -363,7 +364,85 @@ async def proxy(self, host, port, proxied_path):
...
+        else:
+            self.set_status(response.code, response.reason) # Should already have been set
+
+            dump_headers(headers_raw) # Should already have been emptied
+
+            if response.body: # Likewise, should already be chunked out and flushed
+                self.write(response.body)
+
+    async def _proxy_normal(self, host, port, proxied_path, body):
         if self.unix_socket is not None:
```
yuvipanda, line 446: This unix socket check should probably be common to both the proxying methods, and should be moved up.

yuvipanda:
ok, I've done one round of review @ganisback! Hope this helps.

@manics I think we should definitely add this feature! Ideally, I'd have liked us to make *everything* streaming, but that's incompatible with `RewriteableResponse` :( So making at least eventstreams (which by definition need to be streaming) streaming (and hence incompatible with RewriteableResponse) seems ok to me.

ganisback:
OK, will address your comments in the weekend

yuvipanda:
Hey @ganisback! Just wanted to check to see if there's anything else I can do to help you move this forward :) I just ran into this again in a different project!

jupyter_server_proxy/handlers.py
ganisback: Fixed

jupyter_server_proxy/handlers.py
ganisback: fixed

jupyter_server_proxy/handlers.py
ganisback: parse_line is a internal method in HTTPHeaders, we can not use it directly

ganisback:
Hi @yuvipanda , I am blocking in testing case, can you help add the testing case?

jupyter_server_proxy/handlers.py
ganisback: Fixed

yuvipanda:
@ganisback it was slightly tricky, but I think I have a useful test for you at https://github.com/ganisback/jupyter-server-proxy/pull/1! It fails on main but passes on this branch!

LMK what you think of it. I haven't had time to look at your changes yet unfortunately but will do this coming week!

ganisback:
@yuvipanda I increased resp time and added resp data checking, now the tests passed.

jupyter_server_proxy/handlers.py
yuvipanda: Aaah, I see it doesn't return it but updates it. 

The regex here still worries me though. Is it coming from somewhere specific (like an RFC, or an existing implementation?)? If so, I'd appreciate a link to the source as a comment. If not, I'd like us to find some way to farm this out. I would avoid writing code that parses HTTP in any form if possible.

jupyter_server_proxy/handlers.py
```diff
@@ -373,6 +373,82 @@ async def proxy(self, host, port, proxied_path):
...
+            for line in headers_raw:
+                r = re.match('^([a-zA-Z0-9\-_]+)\s*\:\s*([^\r\n]+)[\r\n]*$', line)
+                if r:
+                    k,v = r.groups([1,2])
+                    if k not in ('Content-Length', 'Transfer-Encoding',
+                                  'Content-Encoding', 'Connection'):
+                        # some header appear multiple times, eg 'Set-Cookie'
+                        self.set_header(k,v)
+                else:
+                    r = re.match('^HTTP[^\s]* ([0-9]+)', line)
```
yuvipanda, line 399: Same as my comment in https://github.com/jupyterhub/jupyter-server-proxy/pull/479/files#r1645266065

jupyter_server_proxy/handlers.py
```diff
@@ -373,6 +373,82 @@ async def proxy(self, host, port, proxied_path):
             )
         else:
             client = httpclient.AsyncHTTPClient()
+        # check if the request is stream request
+        proxy_streaming = self.request.headers.get('Accept')
```
yuvipanda, line 377: Let's rename this variable to `accept_header` to more descriptively describe what it is

tests/test_proxies.py
```diff
@@ -343,6 +344,43 @@ def test_server_content_encoding_header(
...
+        stream_read_intervals.append(time_taken)
+        times_called += 1
+        stream_data.append(data)
+
+    url = f"http://{LOCALHOST}:{PORT}/python-eventstream/stream/{limit}?token={TOKEN}"
+    client = AsyncHTTPClient()
+    await client.fetch(
+        url,
+        headers={"Accept": "text/event-stream"},
+        request_timeout=22,
```
yuvipanda, line 374: I think explicitly setting timeout can be removed now right?

tests/test_proxies.py
```diff
@@ -343,6 +344,43 @@ def test_server_content_encoding_header(
...
+        streaming_callback=streaming_cb,
+    )
+    assert times_called == limit
+    print(stream_read_intervals)
+    assert all([0.45 < t < 3.0 for t in stream_read_intervals])
+    print(stream_data)
```
yuvipanda, line 380: Let's get rid of these extra print statements

yuvipanda (changes requested):
Almost there! Can you address https://github.com/jupyterhub/jupyter-server-proxy/pull/479/files#r1645273654 and https://github.com/jupyterhub/jupyter-server-proxy/pull/479/files#r1645272938, and the couple other style things? Then it's ready to land I think.

Thank you for your contribution and patience, @ganisback!

jupyter_server_proxy/handlers.py
ganisback: fixed

tests/test_proxies.py
ganisback: fixed

tests/test_proxies.py
ganisback: fixed

jupyter_server_proxy/handlers.py
ganisback: here is a example for stream header:
```
HTTP/1.1 200 OK
Content-Type: text/html
Cache-Control: no-cache
Connection: keep-alive
```
httputil.HTTPHeaders.parse_line  can not parse the first line to get http status, so we use regex to solve this issue.
the source code comes from https://github.com/ideonate/jhsingle-native-proxy/blob/3e26a4ee0e7318970b7cf6abbd7d88455a9ac621/jhsingle_native_proxy/proxyhandlers.py#L265

ganisback:
@yuvipanda any other comments?

krassowski:
Excited for this! Spent to much of my Saturday debugging why SSE was connected but not sending anything.

krassowski:
@yuvipanda @manics is there anything else that needs to be done here to get it merged?

krassowski:
Hi folks, just circling back here, is there any way I could help with this PR?

ganisback:
> Hi folks, just circling back here, is there any way I could help with this PR?

I'm just waiting for they merge this PR

yuvipanda:
Sorry for dropping this, travel / vacation time :( I'll try to spend an hour on this later this week.

krassowski:
It looks like pre-commit is failing because this PR is behind `main` branch as of now, for example it does not include the `rawsocket.py` file which was added last month (https://github.com/jupyterhub/jupyter-server-proxy/blob/main/jupyter_server_proxy/rawsocket.py)

As per https://results.pre-commit.ci/run/github/71295164/1723882819.1ODn2EWuTr2D6HI5NFt0pQ

image

krassowski:
I opened a PR against this branch (https://github.com/ganisback/jupyter-server-proxy/pull/2) to fix pre-commit checks

yuvipanda:
I still don't like using regexes for parsing headers, but I this is a clear improvement over status quo. So I've rebased this, and will merge. Apologies for the long time this one took, @ganisback! 

I've opened https://github.com/jupyterhub/jupyter-server-proxy/issues/498 to handle removing the regexes.

krassowski:
Amazing, thanks so much! Do you plan to cut a new release with it now or only after https://github.com/jupyterhub/jupyter-server-proxy/issues/498 gets addressed?

yuvipanda:
#498 doesn't need to block a release!

yuvipanda:
If you make a PR following https://github.com/jupyterhub/jupyter-server-proxy/blob/main/RELEASE.md i'm happy to merge that :)

yuvipanda:
Although i'm off from monday!

PR #481, opened 2024-06-25
Fix failure to update __version__

consideRatio:
During our release process, the `_version.py` file wasn't updated as it should been so when inspecting `__version__` we actually just got `4.1.1-0.dev` in the released package where we expected `4.2.0`.


consideRatio:
Tests are unrelated and failing in main branch, I'm self merging this minor fix.

PR #486, opened 2024-06-26
Adding a complete example to clarify where to put the config files.

dragz:
A small addition to the examples. I wasted a few hours trying to figure out where to put the config files. It might save some time for others.

### DIFF
--- a/docs/source/examples.md
+++ b/docs/source/examples.md
@@ -16,7 +16,7 @@ Projects can also add the `jupyter-server-proxy` topic to the GitHub repository 
 # Test the proxy setup
 
 If you just want to test if the proxy config is correct you can put the following into `$HOME/.jupyter/jupyter_server_config.py`:
-```
+```python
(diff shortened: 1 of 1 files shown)


manics:
Thanks, I think having a full example makes sense. What do you think about linking to our test config instead:
https://github.com/jupyterhub/jupyter-server-proxy/blob/main/tests/resources/jupyter_server_config.py

The advantage of that is it's always tested so will never go out of date, and also includes full examples for all config options.

dragz:
I tried copying the mentioned jupyter_server_config.py but it does not work out of the box. First it fails to launch the singleuserserver because it cannot find proxyextension. Commenting those two lines out make the server start but all test examples that shows up gives 500 internal error. All this will confuse a user trying to set this up.

manics:
That's a fair point! Can you change your example so that it'll work on all systems? At the moment it'll only work with anaconda. For example, import and use `sys.executable`

Issue #489, opened 2024-07-01
Enforce pre-commit in this repo


manics:
### Proposed change
We have a [`.pre-commit-config.yaml`](https://github.com/jupyterhub/jupyter-server-proxy/blob/main/.pre-commit-config.yaml) config file in this repo, but we don't enforce it. We should enforce it.

### Alternative options
Delete `.pre-commit-config.yaml` if we're not going to enforce it

### Who would use this feature?
Developers

### (Optional): Suggest a solution
Either activate pre-commit.ci, or add a GH workflow to run https://github.com/pre-commit/action

manics:
I've enabled this repo in pre-commit.ci

PR #492, opened 2024-07-10
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -223,3 +223,10 @@ concurrency = [
 
 [tool.coverage.html]
 show_contexts = true
+
+[tool.codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = '.git*,*.svg,*.lock'
+check-hidden = true
+ignore-regex = '\bTE\b'
+# ignore-words-list = ''
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,6 +69,15 @@ repos:
     hooks:
       - id: flake8
 
+  # Lint: Typos
+  - repo: https://github.com/codespell-project/codespell
+    # Configuration for codespell is in pyproject.toml
+    rev: v2.2.6
+    hooks:
+    - id: codespell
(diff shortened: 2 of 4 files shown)


manics:
Is it possible to use codespell with just pre-commit, so we don't need another GitHub workflow?

yarikoptic:
> Is it possible to use codespell with just pre-commit, so we don't need another GitHub workflow?

yes*!  I just was not sure if any of existing workflows already would run pre-commit on changes -- I pushed TEMP commit with a typo, let's see if gets picked up... so far I do not see any other CI step raising an error...

* codespell workflow also includes "matcher" step which annotates PR diff on where&which typo has happened like did now:

image

if we remove workflow - that would be gone - but it would just remove that convenience.

manics:
We can deal with enforcing pre-commit in https://github.com/jupyterhub/jupyter-server-proxy/issues/489

I think keeping things simple and using just pre-commit is best, especially as typos should be infrequent after an initial review of the repo.

yarikoptic:
Coolio, your call, rebased kicking away action and the TEMP commit with a typo, took out of draft.

pyproject.toml
```diff
@@ -223,3 +223,10 @@ concurrency = [
 
 [tool.coverage.html]
 show_contexts = true
+
+[tool.codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = '.git*,*.svg,*.lock'
+check-hidden = true
+ignore-regex = '\bTE\b'
```
manics, line 231: Can you add a comment for why this is needed (presumably you've found some intentional use of `TE`?)

manics (commented):
Thanks. Ignore the pre-commit failure since it's due to pre-exisitng formatting issues, codespell is passing.

PR #499, opened 2024-08-29
labextension: refreeze yarn.lock

krassowski:
Following https://github.com/jupyterhub/jupyter-server-proxy/blob/main/RELEASE.md to release a new version as per https://github.com/jupyterhub/jupyter-server-proxy/pull/479#issuecomment-2317367188


krassowski:
CC @yuvipanda (I am blocked here as the instructions say "and continue only when its merged.")

yuvipanda:
Hmm, I don't know why it wants to do that but ok!

PR #500, opened 2024-08-29
Add changelog for v4.4.0 (?)

krassowski:
As per [RELEASE](https://github.com/jupyterhub/jupyter-server-proxy/blob/main/RELEASE.md) instructions I run `github-activity jupyterhub/jupyter-server-proxy --since v4.3.0`. That gave a bit different format than the previous changelogs, I think because labels were not enforced. I manually split PRs into appropriate groups and adjusted the format to follow that of previous entries.

I think the next version should be either 4.3.1 or 4.4.0. Conservatively I assumed it will more likely be 4.4.0 but please do change it if it is not correct or just let me know and I will fix it up.

CC @yuvipanda :)

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -8,7 +8,6 @@
 
 - feat: support stream api [#479](https://github.com/jupyterhub/jupyter-server-proxy/pull/479) ([@ganisback](https://github.com/ganisback))
 
-
 #### Bugs fixed
 
 - fix: missing tornadoimport for rawsocket handler [#491](https://github.com/jupyterhub/jupyter-server-proxy/pull/491) ([@imfing](https://github.com/imfing))
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -1,5 +1,32 @@
 # Changelog
 
+## 4.4
+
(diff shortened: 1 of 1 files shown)


yuvipanda:
Sure, let's try it and see :)

What steps are left?

krassowski:
I think just bumping the version with `tbump` and making the release as per https://github.com/jupyterhub/jupyter-server-proxy/blob/main/RELEASE.md - I think I cannot help here as it requires push access

yuvipanda:
Given that you have rights in jupyterlab and similar places, i've invited you to this repo! Welcome :)

krassowski:
:tada: Thank you for the help here, the new version is now on PyPI: https://pypi.org/project/jupyter-server-proxy/4.4.0/ and the version on `main` was reset to dev. The last step is to merge a conda-forge PR once CI passes: https://github.com/conda-forge/jupyter-server-proxy-feedstock/pull/42

yuvipanda:
ty @krassowski. please also test the version from pypi once to make sure it is fine?

krassowski:
Just circling back to confirm that I tested the latest version from conda-forge :)

PR #501, opened 2024-09-10
Integrating native-proxy

jwindgassen:
Hey everyone.

I was recently made aware, of the desire to be able to create standalone proxies, similar to how it is done in [jhsingle-native-proxy](https://github.com/ideonate/jhsingle-native-proxy) (see #1).
This would be immensely advantageous for us, so I started porting the code here recently.

There is still a lot to do and I needed to remove/comment out a few of the original features, but it is already fundamentally working as is. I am opening this PR to let you know of this and get an opinion on a few bits here and there. I will continue to improve on it in the next weeks.

In the meantime, any comments and ideas are welcome :)


ryanlovett:
Hi @jwindgassen , thanks for this contribution!

Is the idea to replace jhsingle-native-proxy (so that code can be shared), or will that project continue on? If the latter then could there still be duplication of code? (will this need to periodically get resynced with jhsingle-native-proxy?) Or should shared code be factored out into something separate for each project to use?

If there is documentation on using this, could that be integrated as well?

jwindgassen:
That is a good question. I do not know about the current state of the original project. The newest commit was already made a few months ago, but I do not think it is abandoned. I will contact the original developer soon and ask him about the current status and his opinion on this.

But the proxies used by it are almost identical to the proxies here, and (I think) they originally were a copy of an old version of the ones in this package.
IMHO, the best would either be completely merging it into here, or making JSP a dependency of `jhsingle-native-proxy` and importing the Proxies there. But that is, of course, not my decision to make but one of the original developer.
Furthermore, `jhsingle-native-proxy` does not support Unix Sockets, which we would love to use.
But since both projects have a lot of similarity, keeping them completely separate from one another is a lot of redundancy.

The Documentation for it is currently in the ReadMe, but could be added to the docs here, given that we decide to add this feature.

aktech:
This sounds like a great idea. I would suggest to create some kind of checklist of the features that are ported from [jhsingle-native-proxy](https://github.com/ideonate/jhsingle-native-proxy) and what's not during the course of this pull request (or a series of pull requests).

We're using [jhsingle-native-proxy](https://github.com/ideonate/jhsingle-native-proxy) heavily in [jhub-apps](https://github.com/nebari-dev/jhub-apps) would love to move to just using `jupyter-server-proxy` for everything we can, and having a feature parity checklist will help us understand, what we'll be having/missing with this port.

jwindgassen:
@ryanlovett I talked to the original developer. He welcomes the merging of his feature into Jupyter Server Proxy. There are currently no further plans for `jhsingle-native-proxy`, but it will persist (for now), as there are a few custom additions, which I currently do not plan to integrate here.

@aktech Sure. Here is a list with the features I have currently ported or plan to do so:
- [x] Launching standalone proxies from CLI
- [x] Make JupyterHub, both as dependency and the required authentication with it, optional
- [x] Automatically enable JupyterHub Authentication when we are spawned by one
- [x] Send Activity Notifications to JupyterHub
- [ ] Ensure usability with Unix-Sockets
- [ ] Allow customization of `environment `and `mappath`
- [ ] Progressive Proxies (#502)
- [ ] Docs & Tests (maybe even with JupyterHub?)
- [ ] (Maybe) Re-add pulling of Git Repository and venv/conda activation?

That's at least everything I can think of now. If you have other ideas or require something more, let me know and I will add it to the list :)

ryanlovett:
@jwindgassen Thanks very much for asking! It sounds like this has the potential to consolidate development in the future.

jwindgassen:
@ryanlovett @aktech I have been working on the feature over the last few weeks, and I am now happy to announce, that the code is working 🎉 
I tested it with my own TLJH instance, and it also works with our production JupyterHub Setup. I also verified that authentication with JupyterHub works.
There are a few small things I will probably improve in the future, but I am quite happy with the state of the code as it is now.

I'm welcoming anyone to test the feature on their JupyterHub instance for testing. Please let me know about any problems or errors you encounter when doing so 🙂

### How to use
For testing, I like to use [voila](https://github.com/voila-dashboards/voila/). The command to execute might look like this:
`jupyter standaloneproxy --debug -- voila --no-browser --port={port} /path/to/notebook.ipynb"`

What else do you need from my side, besides the code, to be satisfied with this PR? I am currently writing a page for the docs, which I will commit soon.
I was also planning to write a few tests for the feature, but I will have to look into that later.

ryanlovett:
@jwindgassen Thanks for the update! I'll try to test this week locally.

How might a hub administrator typically configure use of this feature? For example would they set every user to launch voila via standalone proxy from `c.Spawner.cmd` instead of jupyterhub-singleuser?

Is the intent of standalone to essentially re-use the hub's auth, spawner, user storage, etc. but limit what apps users can invoke because it specifies just the one? (since jupyter server + jupyter-server-proxy enables users to launch an arbitrary number of proxied apps)

jwindgassen:
In essence, yes. 
With the standalone proxy feature, admins/developers of the JupyterHub can give the users of the Hub access to other applications instead of just Notebook/JupyterLab.
While we can currently achieve the same with the current jupyter-server-proxy via the button in the Launcher, this is often an indirection when you only want to access e.g. RStudio. 
By using the `SuperviseAndProxyHandler` directly, without attaching it to a `jupyter-server`, we can (re)use the Authentication, User Model, Proxy, etc. JupyterHub provides. Secondly, it is also probably easier to integrate new web apps into a cloud environment this way, than to manually setup everything yourself.

But since you need to overwrite `c.Spawner.cmd` or modify the behavior of the Spawner in a subclass, new Apps can only be added by the administrators. This gives users access to different apps on the JupyterHub landing page, similar to [Open OnDemand](https://openondemand.org/run-open-ondemand). JupyterHub then needs to switch between the launch commands accordingly.
In our case, we are planning to provide `RStudio`, `XpraHTML` (Remote Desktop), `MATLAB`, `NEST-Desktop`, and more to come in the future, using this feature.

ryanlovett:
Fascinating, thanks!

More of an aside, but how are you customizing the spawner to launch the different applications?

jwindgassen:
No. We have created our own custom Spawner. But is similar to this. We have overwritten the start method, which will submit a new Job to our cluster. And inside the start script, we start jupyterhub singleuser at the end.

But now you mention it, jhub-apps might synergyze quite well with the standalone feature.

jupyter_server_proxy/standalone/__init__.py
```diff
@@ -0,0 +1,214 @@
...
+        "--websocket-max-message-size",
+        default=0,
+        type=int,
+        help="Max size of websocket data (leave at 0 for library defaults).",
+    )
+    parser.add_argument(
+        "command", nargs="+", help="The command executed for starting the WebApp"
+    )
+
+    args = parser.parse_args()
```
manics, line 209: There's a mismatch between the options supported by Jupyter Server Proxy
https://github.com/jupyterhub/jupyter-server-proxy/blob/v4.4.0/jupyter_server_proxy/config.py#L34-L47
and those available here. Obviously some options aren't relevant, but for example `raw_socket_proxy` is.

jupyter_server_proxy/standalone/__init__.py
```diff
@@ -0,0 +1,214 @@
...
+        description="Wrap an arbitrary WebApp so it can be used in place of 'jupyterhub-singleuser' in a JupyterHub setting.",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+
+    parser.add_argument(
+        "-p",
+        "--port",
+        type=int,
+        dest="port",
+        help="Set port for the proxy server to listen on. Will use 'JUPYTERHUB_SERVICE_URL' or '127.0.0.1' by default.",
```
manics, line 115: The help messages for `--port` and `--address` are the wrong way round.

manics (commented):
The command line arguments are quite complicated, e.g. having to parse maps.

The alternative is more work, but if we were to refactor ServerProcess to be a Traitlets Configurable
https://github.com/jupyterhub/jupyter-server-proxy/blob/41e588ab3dfb3db5bdf94c151eaa7ff308d94571/jupyter_server_proxy/config.py#L31-L49
I think we could take advantage of Traitlets ability to automatically create all CLI args, and we'd have the benefit of being able to configure jupyter-standaloneproxy using an arbitrarily complex file.

What do you think? I'm happy to investigate converting ServerProxy to Traitlets.

jwindgassen:
Thanks for the suggestion, I really like that idea. This should also solve the issue with keeping CLI Arguments up with any new options added to the proxies. I will look into it :)

manics:
I've made a start in https://github.com/jupyterhub/jupyter-server-proxy/pull/507

jwindgassen:
@manics

So I used your branch to create the CLI via traitlets. I'm no expert with traitlets, but I managed to get it working: https://github.com/jwindgassen/jupyter-server-proxy/commit/2d9eb5bb593e4a4f4a7b535ffc831ece36ea6548

I had to play around with aliases and the extra_args a bit, but I wanted to keep the CLI reasonably unchanged to before.
Getting the command, which was previously a positional argument, was a bit tricky, since I couldn't find a proper way to add a positional argument to the Parser traitlets create. It was possible by overwriting `_create_loader()`, but not really pretty. Using `extra_args` instead works, but we now lack a proper explanation of command when using `--help`.

If you have a better idea or otherwise comments on my changes over there, let me know! :)

manics:
Sorry for the delay. Since this is a new addition to Jupyter server proxy I think we should prioritise long term maintenance over just replicating the previous CLI- if there's a better way to do things we can use it as an opportunity to refactor.

We also don't need to do everything in one go, for example it's fine to initially focus on creating a functional standalone proxy along that only supports standard traitlets configuration, and add additional flags in a follow-up PR.

jwindgassen:
I'm fine with how the CLI looks right now, so I'm happy to switch to this once your PR has been accepted.

I am also almost done with Tests and Docs, they should be ready by the end of the week.

jwindgassen:
Ok, I have now also added Docs and Tests for the new feature.

Writing proper tests for Login and Activity is a bit more difficult, since I would need to spawn a JupyterHub instance to gain full access to its API. For now, I limited it to only testing our code, since the classes I import from JupyterHub are tested over there.

I also added a section to the docs, mostly targeted to developers, where I explain the different sections of the code and what features I needed to implement to make this work smoothly.

If you think we need more tests for specific cases or want something to be explained in the docs in more detail, let me know.

jhgoebbert:
@jwindgassen So from your perspective this is ready to be merged. Great and Thank you!

( As soon as https://github.com/jupyterhub/jupyter-server-proxy/pull/507 and https://github.com/jupyterhub/jupyter-server-proxy/pull/508 is merged this native-proxy can be updated afterwards. But for now this PR here is implemented to be independent of them. True? )

jwindgassen:
Yes. This is currently independent of #507 and functions without it. But in the future, once that has been merged, I would update the standalone feature to use `traitlets.Configurable`. It would be a lot tidier and future-proof using that approach.

@aktech Have you been able to get it running and did it work in your setup? I would highly appreciate any feedback or comments on this 🙂 

@ryanlovett How do we continue for this PR? Is there anyone specific responsible for reviewing it? Is there still more you need here? Sorry for being a bit impatient, but we would like to get this feature running soon on our JupyterHub instance.

ryanlovett (changes requested):
I've left a few minor comments and suggested changes. What do you think?

ryanlovett:
In terms of next steps, I'm fine with merging if @manics thinks its okay to go ahead and then address the related PRs later.

jupyter_server_proxy/standalone/__init__.py
```diff
@@ -0,0 +1,216 @@
...
+from tornado import ioloop
+from tornado.httpserver import HTTPServer
+from tornado.log import app_log as log
+from tornado.log import enable_pretty_logging, gen_log
+
+from .activity import start_activity_update
+from .proxy import configure_ssl, make_proxy_app
+
+
+def _default_address_and_port() -> tuple[str, int]:
```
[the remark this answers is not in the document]
jwindgassen: jupyter-server-proxy officially supports Python >= 3.8. But while that type annotation only works since 3.9, we can use `from __future__ import annotations` to enable it in older versions (back until 3.7). But I will make the comment more specific.

aktech (changes requested):
@jwindgassen Thanks a lot for working on this. This is really very useful. I tried replacing [jhsingle-native-proxy](https://github.com/ideonate/jhsingle-native-proxy) with this branch and I was able to get it working partially (minus conda env and repo) for panel and voila apps. 

To be able to completely use it, we would need ability to specify conda env and pull from repo, but I think this PR is a great start in that direction. These features can be contributed in later PRs and this PR looks complete enough to get merged IMO.

I tested for panel and voila. For panel I ran this:

```
jupyter standaloneproxy --debug --skip-authentication -- python -m bokeh_root_cmd.main ~/path/to/jhub-apps/jhub_apps/examples/panel_basic.py --port={port} --debug --allow-websocket-origin=127.0.0.1:8888 --server=panel
```

manics:
@ryanlovett If you're generally happy with this would you mind merging https://github.com/jupyterhub/jupyter-server-proxy/pull/507 first, then we can rebase this PR, and it'll make this PR smaller, and it'll be a lot clearer what we're adding.

@aktech I don't think we should completely reproduce jhsingle-native-proxy since we have to maintain this long term- cloning git repos and setting up conda envs doesn't feel like it's in scope. However I'd hope the move to traitlets makes it much easier to extend this in a separate app, or perhaps it's as easy as wrapping it in another script?

jwindgassen:
@ryanlovett @manics Now that #507 is merged, should I rebase and tidy up the commits here? Or should I merge main into here and then append the required changes to the CLI?

manics:
Up to you! I usually rebase my PRs.

jwindgassen:
Ok. So I rebased on top of the new #507 Merge.

I had to refactor the `ServerProcess` Configurable a bit more to reuse it for the standalone. However, when done like this, it should reasonably well future-proof for any new attributes and functionality that might be added.

There are still 2 minor changes I am thinking about implementing:
- Instead of using `traitlets.config.Application`, we could use `jupyter_core.application.JupyterApp`. This adds the `--config` argument, meaning existing JupyterHub setups could reuse their `config.py` file. A `config.py` might also be a bit nicer when setting up the standalone proxy for complicated setups
- As I mentioned at the end of #507, I think we should be able to make `ServerProcess` and `LauncherEntry` a `HasTraits` instead of a `Configurable`.

Any comments, on the changes or these ideas, are very welcome :) @manics @ryanlovett

jhgoebbert:
This looks great to me. Can it be merged?

jupyter_server_proxy/standalone/app.py
```diff
@@ -0,0 +1,289 @@
...
+                return 443
+
+        return 8888
+
+    server_port = Int(default_value=0, help=ServerProcess.port.help).tag(config=True)
+
+    activity_interval = Int(
+        default_value=300,
+        help="""
+        Specify an interval to send regulat activity updated to the JupyterHub (in Seconds).
```
manics, line 107:
```suggestion
        Specify an interval to send regular activity updated to the JupyterHub (in seconds). 
```

jupyter_server_proxy/standalone/app.py
```diff
@@ -0,0 +1,289 @@
...
+                dedent(ServerProcess.raw_socket_proxy.help),
+            ),
+            "skip-authentication": (
+                {"StandaloneProxyServer": {"skip_authentication": True}},
+                dedent(self.__class__.skip_authentication.help),
+            ),
+        }
+
+        # Create an Alias to all Traits defined in ServerProcess, with some
+        # exeptions we do not need, for easier use of the CLI
```
manics, line 149:
```suggestion
        # exceptions we do not need, for easier use of the CLI
```

jupyter_server_proxy/standalone/proxy.py
```diff
@@ -0,0 +1,81 @@
...
+            return set()
+
+        def set_default_headers(self):
+            self.set_header("X-JupyterHub-Version", __jh_version__)
+
+        def prepare(self, *args, **kwargs):
+            pass
+
+        def check_origin(self, origin: str = None):
+            # Skip JupyterHandler.check_origin
```
manics, line 60: Can you explain why it's OK to go straight to th WebSocketHandler check- is it because
https://jupyter-server.readthedocs.io/en/stable/api/jupyter_server.base.html#jupyter_server.base.handlers.JupyterHandler.check_origin
implies we're skipping the more relaxed check?

jupyter_server_proxy/standalone/proxy.py
```diff
@@ -0,0 +1,81 @@
...
+            return WebSocketHandler.check_origin(self, origin)
+
+        def check_xsrf_cookie(self):
+            # Skip HubAuthenticated.check_xsrf_cookie
```
manics, line 64: Can you explain why this is safe (assuming it is!)?

manics (commented):
Sorry for the delay in reviewing this, it generally looks good to me, though I haven't yet managed to run it locally.

manics (commented):
The shortened command line arguments have a mix of `-` and `_`, could we use `-` for all?

<details>

```
Wrap an arbitrary web service so it can be used in place of 'jupyterhub-
singleuser'  in a JupyterHub setting.

Usage: jupyter standaloneproxy [options] -- <command>

For more details, see the jupyter-server-proxy documentation.

Options
=======
The options below are convenience aliases to configurable class-options,
as listed in the "Equivalent to" description-line of the aliases.
To see all configurable class-options for some <cmd>, use:
    <cmd> --help-all

--debug
    set log level to logging.DEBUG (maximize logging output)
    Equivalent to: [--Application.log_level=10]
--show-config
    Show the application's configuration (human-readable format)
    Equivalent to: [--Application.show_config=True]
--show-config-json
    Show the application's configuration (json format)
    Equivalent to: [--Application.show_config_json=True]
--generate-config
    generate default config file
    Equivalent to: [--JupyterApp.generate_config=True]
-y
    Answer yes to any questions instead of prompting.
    Equivalent to: [--JupyterApp.answer_yes=True]
--absolute-url
    Proxy requests default to being rewritten to ``/``. If this is True,
    the absolute URL will be sent to the backend instead.
    Equivalent to: [--ServerProcess.absolute_url=True]
--raw-socket-proxy
    Proxy websocket requests as a raw TCP (or unix socket) stream.
    In this mode, only websockets are handled, and messages are sent to the backend,
    similar to running a websockify layer (https://github.com/novnc/websockify).
    All other HTTP requests return 405 (and thus this will also bypass rewrite_response).
    Equivalent to: [--ServerProcess.raw_socket_proxy=True]
--skip-authentication
    Do not authenticate access to the server via JupyterHub. When set,
    incoming requests will not be authenticated and anyone can access the
    application.

    WARNING: Disabling Authentication can be a major security issue.
    Equivalent to: [--StandaloneProxyServer.skip_authentication=True]
--log-level=<Enum>
    Set the log level by value or name.
    Choices: any of [0, 10, 20, 30, 40, 50, 'DEBUG', 'INFO', 'WARN', 'ERROR', 'CRITICAL']
    Default: 30
    Equivalent to: [--Application.log_level]
--config=<Unicode>
    Full path of a config file.
    Default: ''
    Equivalent to: [--JupyterApp.config_file]
--absolute_url=<Bool>
    Proxy requests default to being rewritten to ``/``. If this is True, the
    absolute URL will be sent to the backend instead.
    Default: False
    Equivalent to: [--StandaloneProxyServer.absolute_url]
--environment=<Union>
    A dictionary of environment variable mappings. As with the command traitlet,
    ``{port}``, ``{unix_socket}`` and ``{base_url}`` will be substituted.
    Could also be a callable. It should return a dictionary.
    Default: {}
    Equivalent to: [--StandaloneProxyServer.environment]
--mappath=<Union>
    Map request paths to proxied paths. Either a dictionary of request paths to
    proxied paths, or a callable that takes parameter ``path`` and returns the
    proxied path.
    Default: {}
    Equivalent to: [--StandaloneProxyServer.mappath]
--port=<Int>
    The port where the proxy server can be accessed. The port is usually taken
    from the `JUPYTERHUB_SERVICE_URL` environment variable or will default to
    `8888`. Used to explicitely overwrite the port of the server.
    Default: 0
    Equivalent to: [--StandaloneProxyServer.port]
--raw_socket_proxy=<Bool>
    Proxy websocket requests as a raw TCP (or unix socket) stream. In this mode,
    only websockets are handled, and messages are sent to the backend, similar
    to running a websockify layer (https://github.com/novnc/websockify). All
    other HTTP requests return 405 (and thus this will also bypass
    rewrite_response).
    Default: False
    Equivalent to: [--StandaloneProxyServer.raw_socket_proxy]
--request_headers_override=<Union>
    A dictionary of additional HTTP headers for the proxy request. As with the
    command traitlet, ``{port}``, ``{unix_socket}`` and ``{base_url}`` will be
    substituted.
    Default: {}
    Equivalent to: [--StandaloneProxyServer.request_headers_override]
--timeout=<Int>
    Timeout in seconds for the process to become ready, default 5s.
    Default: 5
    Equivalent to: [--StandaloneProxyServer.timeout]
--unix_socket=<Union>
    If set, the service will listen on a Unix socket instead of a TCP port. Set
    to True to use a socket in a new temporary folder, or a string path to a
    socket. This overrides port.
    Proxying websockets over a Unix socket requires Tornado >= 6.3.
    Default: None
    Equivalent to: [--StandaloneProxyServer.unix_socket]
--base_url=<Unicode>
    Base URL where Requests will be received and proxied. Usually taken from the
    "JUPYTERHUB_SERVICE_PREFIX" environment variable (or "/" when not set).  Set
    to overwrite.
    When setting to "/foo/bar", only incoming requests starting with this prefix
    will be answered by the server and proxied to the proxied app. Any other
    requests will get a 404 response.
    Default: ''
    Equivalent to: [--StandaloneProxyServer.base_url]
--address=<Unicode>
    The address where the proxy server can be accessed. The address is usually
    taken from the `JUPYTERHUB_SERVICE_URL` environment variable or will default
    to `127.0.0.1`. Used to explicitely overwrite the address of the server.
    Default: ''
    Equivalent to: [--StandaloneProxyServer.address]
--server_port=<Int>
    Set the port that the service will listen on. The default is to
    automatically select an unused port.
    Default: 0
    Equivalent to: [--StandaloneProxyServer.server_port]
--activity_interval=<Int>
    Specify an interval to send regulat activity updated to the JupyterHub (in
    Seconds).  When enabled, the StandaloneProxy will try to send a POST request
    to the JupyterHub API containing a timestamp and the name of the server. The
    URL for the activity Endpoint needs to be specified in the
    "JUPYTERHUB_ACTIVITY_URL" environment variable. This URL usually is
    "/api/users/<user>/activity".
    Set to 0 to disable activity notifications.
    Default: 300
    Equivalent to: [--StandaloneProxyServer.activity_interval]
--websocket_max_message_size=<Int>
    Restrict the size of a message in a WebSocket connection (in bytes). Tornado
    defaults to 10MiB.
    Default: None
    Equivalent to: [--StandaloneProxyServer.websocket_max_message_size]

command=<Union>
    An optional list of strings that should be the full command to be executed.
    The optional template arguments ``{port}``, ``{unix_socket}`` and
    ``{base_url}`` will be substituted with the port or Unix socket path the
    process should listen on and the base-url of the notebook.
    Could also be a callable. It should return a list.
    If the command is not specified or is an empty list, the server process is
    assumed to be started ahead of time and already available to be proxied to.
    Default: traitlets.Undefined

Examples
--------

    jupyter standaloneproxy -- voila --port={port} --no-browser /path/to/notebook.ipynb

To see all available configurables, use `--help-all`.
```

</details>

There are also three JupyterApp flags:
```
--generate-config
    generate default config file
    Equivalent to: [--JupyterApp.generate_config=True]
-y
    Answer yes to any questions instead of prompting.
    Equivalent to: [--JupyterApp.answer_yes=True]
--config=<Unicode>
    Full path of a config file.
    Default: ''
    Equivalent to: [--JupyterApp.config_file]
```
`-y` Is there any use for this? Otherwise add it to the exclusions.
`--generate-config` doesn't work - it just starts the proxy. Easiest to add to the exclusions unless you think it's useful?
`--config` silently continues if the config file is broken or non-existent. I think this is the default behaviour for Jupyter Apps but I don't think it's very helpful, especially for an app like this where the config is critical!

jwindgassen:
@manics Thank you for the review. 

Both `check_origin` and `check_xsrf_cookie` were not called in the original *jhsingle-native-proxy* implementation, so I assume they are not strictly necessary. The original implementation did not inherit from JupyterHandler, so there was no need to explicitly circumvent it there. We have also overridden `check_xsrf_cookie` in the `ProxyHandler`, mentioning to defer the checking of XSRF to the proxied app. The original implementation inherits from `HubOAuthenticated` like this: `ProxyHandler(HubOAuthenticated, WebSocketHandlerMixin)`, while I inherit like this: `StandaloneHubProxyHandler(HubOAuthenticated, SuperviseAndProxyHandler)`. So I need to override it again.
But I will try to replicate the error I got when I did not skip them and take a closer look, if it is safe to skip them. Otherwise I will expand the comments there.

Regarding the `--generate-config` and `--config` flags, I would like to get those working. For more complicated setups, we will probably need a callable mappath in the future. This is only possible in the config.py. But I will try to ensure the server only runs when the config is loaded correctly.
I will need to rework the CLI anyways when we merge #521, so I will unify and fix it after that.

We are currently working on getting the standalone feature running on our proper JupyterHub instance. We have already found (and fixed) a few bugs in the last few days while doing that. So if it gives you more confidence, we can wait with merging this PR until the standalone feature is fully running on our hub to make sure it is working there :)

jhgoebbert:
Great!
We am able to run
- [Xpra](https://github.com/Xpra-org/xpra) using [jupyter-xprahtml5-proxy](https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy) and also
- VSCode over [OpenVSCode-Server](https://github.com/gitpod-io/openvscode-server) using [jupyter-openvscodeserver-proxy](https://github.com/FZJ-JSC/jupyter-openvscodeserver-proxy)

with this StandaloneProxyServer directly from our JupyterHub.

jhgoebbert:
For Xpra the start command is:
`jupyter-standaloneproxy --config=$HOME/.jupyter/standalone-xpra-config.py`
with an `standalone-xpra-config.py` ==
```
import logging
import os
from jupyter_xprahtml5_proxy import setup_xprahtml5

# Execute the setup from the Jupyter proxy to retrieve the startup information
server_config = setup_xprahtml5()

# Enable debugging and use the correct network address
c.StandaloneProxyServer.log_level = logging.DEBUG
c.StandaloneProxyServer.address = os.environ.get("HOSTNAME", "127.0.0.1")

# Forward the server configuration from the Jupyter proxy to the standalone app
c.StandaloneProxyServer.merge(server_config)
```

For OpenVSCode-Server the start command is:
`jupyter-standaloneproxy --config=$HOME/.jupyter/standalone-vscode-config.py`
with an `standalone-vscode-config.py` ==
```
import logging
import os
from urllib.parse import urlparse, parse_qs
from jupyter_openvscodeserver_proxy import setup_openvscodeserver

# Execute the setup from the Jupyter proxy to retrieve the startup information
server_config = setup_openvscodeserver()

# Enable debugging and use the correct network address
c.StandaloneProxyServer.log_level = logging.DEBUG
c.StandaloneProxyServer.address = os.environ.get("HOSTNAME", "127.0.0.1")

# Forward the server configuration from the Jupyter proxy to the standalone app
c.StandaloneProxyServer.merge(server_config)

# Forward query params from launcher_entry.path_info, except the token
# The token will be directly added as a cookie to any request made to the server
# We can't add the token in mappath, otherwise we will end up in an infinite redirect loop
url = urlparse(server_config["launcher_entry"]["path_info"])
query_args = parse_qs(url.query)
token = query_args.pop("tkn")[0]
proxy_query_args = "&".join([f"{key}={value[0]}" for key, value in query_args.items()])

c.StandaloneProxyServer.mappath = {
    "/": f"/?{proxy_query_args}"
}

c.StandaloneProxyServer.request_headers_override = {
    "Cookie": f"vscode-tkn={token};"
}
```

jhgoebbert:
@manics It would be great if this patch could be merged.
I can confirm that it works on our side.

manics:
> We are currently working on getting the standalone feature running on our proper JupyterHub instance. We have already found (and fixed) a few bugs in the last few days while doing that. So if it gives you more confidence, we can wait with merging this PR until the standalone feature is fully running on our hub to make sure it is working there :)

@jwindgassen have you pushed all the fixes you're planning to make?

jupyter_server_proxy/standalone/app.py
```diff
@@ -0,0 +1,289 @@
...
+
+    @default("prefix")
+    def _default_prefix(self):
+        return os.environ.get("JUPYTERHUB_SERVICE_PREFIX", "/").removesuffix("/")
+
+    @validate("prefix")
+    def _validate_prefix(self, proposal):
+        return proposal["value"].removesuffix("/")
+
+    skip_authentication = Bool(
```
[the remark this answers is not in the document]
jwindgassen: Let's do that in #521. I will need to refactor the traitlets and add the `tag(config=True)`s there anyway.

jupyter_server_proxy/standalone/proxy.py
jwindgassen: Like I mentioned below, this is also done in the original implementation in `jhsingle-native-proxy` [here](https://github.com/ideonate/jhsingle-native-proxy/blob/master/jhsingle_native_proxy/proxyhandlers.py#L523) and in our implementation [here](https://github.com/jupyterhub/jupyter-server-proxy/blob/main/jupyter_server_proxy/handlers.py#L651).

I need to remove it here again because my inheritance order is different from the original one. There, the `ProxyHandler` is inherited from `HubOAuthenticated` directly and then overwritten, whereas I can only inherit from `HubOAuthenticated` in `StandaloneHubProxyHandler` and need to skip it again.

This does not mean that this is inherently safe. But I trust it enough for now. But we are currently planning to let the whole standalone server get checked by someone with more websec experience than me.
I think this is safe for now, but I will report back when we find it is not!

jupyter_server_proxy/standalone/proxy.py
jwindgassen: The problem with JupyterHandler.check_origin is that it requires an IdentityProvider on the server [to work correctly](https://github.com/jupyter-server/jupyter_server/blob/main/jupyter_server/base/handlers.py#L255). This is currently not the case.

I looked a bit into the whole `IdentityProvider` structure, and we could integrate one into the `StandaloneProxyServer`. Additionally, JupyterHub provides a `JupyterHubIdentityProvider`, which can be used to add the Handler for the `oauth_callback/` route automatically, while I am currently adding it manually.
But integrating it now would probably mean another restructuring and a lot of new testing.

tl;dr: I think it might be a good idea in the future to add an `IdentityProvider` to the `StandaloneProxyServer` and remove this skip, but it's not required right now. 

P.S.: This is also skipped in `jhsingle-native-proxy`, since they removed the `JupyterHandler` inheritance from `ProxyHandler`

jwindgassen:
> @jwindgassen have you pushed all the fixes you're planning to make?
@manics 

At least for this PR, I am done, yes. Take a look at comments on the remaining open code reviews. IMHO, this is ready to be merged. There are still some minor things to finish (like #521), but not in here.

jhgoebbert:
@manics Any chance this can be merged?

PR #507, opened 2024-10-24
Make config.ServerProcess into a Configurable

manics:
This isn't useful in the jupyter-server-extension, but it should be useful in
https://github.com/jupyterhub/jupyter-server-proxy/pull/501#pullrequestreview-2393608506

An added benefit is this makes it easier to add more parameters in future since the declaration of the class and the handling of defaults is now in the same place.

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -3,6 +3,7 @@ Traitlets based configuration for jupyter_server_proxy
 """
 
 import sys
+from textwrap import dedent, indent
 from warnings import warn
 
 if sys.version_info < (3, 10):  # pragma: no cover
@@ -344,6 +345,16 @@ def make_server_process(name, server_process_config, serverproxy_config):
     return ServerProcess(name=name, **server_process_config)
 
 
+def _serverproxy_servers_help():
+    serverprocess_help = ""
+    for k, v in ServerProcess.class_traits().items():
+        help = v.metadata.get("help", "").lstrip("\n").rstrip()
+        if help:
+            help = indent(dedent(help), "    ")
+            serverprocess_help += f"{k}\n{help}\n\n"
+    return serverprocess_help
+
+
 class ServerProxy(Configurable):
     servers = Dict(
         {},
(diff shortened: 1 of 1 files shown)


jwindgassen:
While at it, we should probably also convert `LauncherEntry`. Feel free to copy it from [my implementation](https://github.com/jwindgassen/jupyter-server-proxy/blob/server_process_traitlets/jupyter_server_proxy/config.py#39)

manics:
Thanks, I had LauncherEntry in my local branch, but was hoping to minimise changes. As it happens making it Configurable is the easiest way to set the defaults.

Assuming `ServerProcess` can be configured externally I don't think LauncherEntry needs to be, it can be initialised from a dict as at present.

manics:
A nice benefit of this is we can autogenerate some of the manually written docs:
https://github.com/jupyterhub/jupyter-server-proxy/pull/508

manics:
@yuvipanda what do you think of this?

ryanlovett:
@manics I like how the new class relieves us of having to manually specify ServerProcess' fields. The docs generation is also nice.

Will merge [per request](https://github.com/jupyterhub/jupyter-server-proxy/pull/501#issuecomment-2517683153).

jhgoebbert:
With this patch will break multiple server proxies as is not possible anymore to add a function as `command`:
```
[W 2025-02-24 19:28:56.126 ServerApp] jupyter_server_proxy | extension failed loading with message: TraitError("The 'command' trait of a ServerProcess instance expected a list, not the function '_get_cmd'.")
```

Here are some examples on how `command` is used in a server-proxy
- https://github.com/jhgoebbert/jupyter-rsession-proxy/blob/main/jupyter_rsession_proxy/__init__.py#L206
- https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy/blob/main/jupyter_xprahtml5_proxy/__init__.py#L91
- https://github.com/jhgoebbert/jupyter-nestdesktop-proxy/blob/main/jupyter_nestdesktop_proxy/__init__.py#L130

Issue #502, opened 2024-09-16
Responsive Proxy for other Stream Types


jwindgassen:
With #479, there is now support for proxying a `text/event-stream`, but not any other stream type (video/data). From what I can tell, this was originally not possible because of the way `RewritableResponse` works.

But with the new Stream Support, I don't see `RewritableResponse` used anywhere in the [new implementation](https://github.com/jupyterhub/jupyter-server-proxy/blob/main/jupyter_server_proxy/handlers.py#L386). Furthermore, the reference implementation [ideonate/jhsingle-native-proxy/jhsingle_native_proxy/proxyhandlers.py#L249](https://github.com/ideonate/jhsingle-native-proxy/blob/master/jhsingle_native_proxy/proxyhandlers.py#L249) has no restriction to the stream type and just exposes this as a configurable flag.

So is there still any reason we currently only support `text/event-stream`, or could this be expanded?

jwindgassen:
Hey @yuvipanda and @ganisback

As you were responsible for integrating this feature, I was hoping you might be able to give some more details about this issue.

Issue #494, opened 2024-07-19
second-by-second stuttering instead of continuous data stream


jhgoebbert:
### Bug description

It seems as if a **continuous data stream** of a web application is **forwarded in larger data chunks** when using jupyter-server-proxy. This is a major problem for some web services that cannot live with high latency.

#### Actual behaviour

The effect can be seen in the server proxy [jupyter-xprahtml5](https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy) for the html5 web client of the remote desktop software Xpra, where the video stream stalls significantly every second (see below) - unfortunately, this makes it not a proper remote desktop solution.
![xpra_through-jupyter-server-proxy](https://github.com/user-attachments/assets/7407170f-5798-4f9d-9feb-4af02ecdccc6)

#### Expected behaviour

The assumption that it is jupyter-server-proxy which introduces this second-by-second stuttering can be seen in the comparison with a stand-alone Xpra-setup (on the same machine). Here, the html5 web client only communicates via an SSH tunnel with the same configured Xpra server without jupyter-server-proxy in between. The video playback is smooth and does not stutter.
![xpra_direct-ssh-tunnel](https://github.com/user-attachments/assets/6952de0b-8b02-4549-b4b4-6c0be0fd1d08)

### Personal set up

I am running JupyterLab 4.2.1 with jupyter-server-proxy 4.1.2 on a Rocky Linux 8 cluster.
The laptop is connected through an ssh tunnel and runs Firefox (also tested with Chrome) on Windows 11 (other users see this phenomena also on Linux clients).

More details on the installation of JupyterLab can be found here:
https://github.com/easybuilders/JSC/blob/2024/Golden_Repo/j/Jupyter-bundle/Jupyter-bundle-20240520-GCCcore-12.3.0.eb

manics:
Unfortunately it's not a straightforward fix:
https://github.com/jupyterhub/jupyter-server-proxy/pull/479#issuecomment-2177338362

jhgoebbert:
@manics That fix looks promising! Great!

jhgoebbert:
Many thanks to @jwindgassen.  
The cause of the stuttering shown above is a Prometheus metrics of jupyter-resouce-usage which blocks tornado's main thread.
https://github.com/jupyter-server/jupyter-resource-usage/issues/123

Setting `c.ResourceUseDisplay.enable_prometheus_metrics = False` fixes the problem.

PR #512, opened 2024-11-25
Enable progressive proxy via flag

jwindgassen:
With the current implementation for progressive proxies from #479, we only make the request progressive if the Accept Header of the request is exactly `text/event-stream`. We would like to use this feature with other proxies, where the Mime-Type is not so directly determinable, like with *xpra* (Remote Destop).
As far as I can tell, there is no particular reason to limit the progressive feature to only this use case (see #502).

I changed the code to instead allow configuring the behavior of the proxy via the entrypoint of a proxy. The dictionary returned by the entry point can now contain a value "progressive". When set to either `True` or `False`, we can explicitly enable or disable the progressive proxying. When left empty, we will automatically determine what type of proxy to use (see the `progressive` property).
It can also be a function, when the developer wants to dynamically switch between progressive and buffered proxying.

Note: Merging this will collide with #507 & #508 (and possibly #501)!

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -151,7 +151,7 @@ def make_server_process(name, server_process_config, serverproxy_config):
             "rewrite_response",
             tuple(),
         ),
-        progressive=server_process_config.get("progressive", False),
+        progressive=server_process_config.get("progressive", None),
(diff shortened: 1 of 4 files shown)


jhgoebbert:
Great. This might also fix the issue described in https://github.com/jupyterhub/jupyter-server-proxy/issues/494

PR #519, opened 2025-02-25
Allow ServerProcess.command to be a callable

### LINKED ISSUE #518: callable arguments fail with TraitError(" .. instance expected a list, not the function")
The pull request https://github.com/jupyterhub/jupyter-server-proxy/pull/507 has introduced an issue if `command` is passing a callable function.
```
[W 2025-02-24 19:28:56.126 ServerApp] jupyter_server_proxy | extension failed loading with message: TraitError("The 'command' trait of a ServerProcess instance expected a list, not the function '_get_cmd'.")
```

This breaks multiple server proxies as it is not possible anymore to pass a function using `command`.

Looking at the documentation of `jupyter-server-proxy` it should be possible to pass a callable function
- https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#callable-arguments
    - https://jupyter-server-proxy.readthedocs.io/en/latest/server-process.html#command

Here are some examples on how and why for example `command` is used in server-proxies:
- https://github.com/jhgoebbert/jupyter-rsession-proxy/blob/main/jupyter_rsession_proxy/__init__.py#L206
- https://github.com/FZJ-JSC/jupyter-xprahtml5-proxy/blob/main/jupyter_xprahtml5_proxy/__init__.py#L91
- https://github.com/jhgoebbert/jupyter-nestdesktop-proxy/blob/main/jupyter_nestdesktop_proxy/__init__.py#L130

jwindgassen:
Resolves #518

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -79,8 +79,8 @@ class LauncherEntry(Configurable):
 class ServerProcess(Configurable):
     name = Unicode(help="Name of the server").tag(config=True)
 
-    command = List(
-        Unicode(),
+    command = Union(
+        [List(Unicode()), Callable()],
         help="""
         An optional list of strings that should be the full command to be executed.
         The optional template arguments ``{port}``, ``{unix_socket}`` and ``{base_url}``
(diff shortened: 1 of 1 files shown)


manics (approved):
Thanks! I'll merge this now, but if you have time a test would make sure this mistake doesn't happen in future.

PR #520, opened 2025-02-25
Improve Tests

jwindgassen:
As requested in #519, I added some additional tests, so that traits that accept a callable (command, environment, mappath, etc.) will be tested accordingly.

I might add more test cases, so let me know of anything else.


manics (approved):
Thanks, this LGTM!

PR #526, opened 2025-04-13
Fix websocket requests not mapped via mappath

### LINKED ISSUE #525: Websocket requests are not mapped via mappath
### Bug description

Websocket requests are not mapped via `mappath`. The function is not even called on the websocket URL.

### How to reproduce

Write a small server proxy that contains a websocket request, use the `mappath` config argument to define a function that e.g. just prints and returns the current path. It is not called for the websocket request, the console shows

```
Trying to establish websocket connection to ws://...
Websocket connection established to ws://...
```

where `ws://...` is the original path, not mapped using `mappath`.

#### Expected behaviour

`ws://...` should be a path mapped using the custom `mappath` function.

#### Actual behaviour

`ws://...` is the original, unmapped path.

### Your personal set up

 - OS: Fedora Workstation 41
 
 - Version(s): Jupyter Server Proxy version 4.4.0

2xB:
This is done by instead of WebSocketHandlerMixin providing the default `get` method and delegating to `ProxyHandler.http_get`, it now provides `get_websocket` and `ProxyHandler.proxy` delegates to it.

Code is also removed that was never reached since it checked for the "Upgrade" HTTP header after removing it among others defined in `hop_by_hop_headers`.

Fixes #525

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -16,7 +16,7 @@ import aiohttp
 from jupyter_server.base.handlers import JupyterHandler, utcnow
 from jupyter_server.utils import ensure_async, url_path_join
 from simpervisor import SupervisedProcess
-from tornado import httpclient, httputil, web, websocket
+from tornado import httpclient, httputil, web
 from tornado.simple_httpclient import SimpleAsyncHTTPClient
 from traitlets import Bytes, Dict, Instance, Integer, Unicode, Union, default, observe
 from traitlets.traitlets import HasTraits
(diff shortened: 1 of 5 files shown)


2xB:
I believe tests may fail thanks to the custom `NewHandler` implemented for the tests in https://github.com/jupyterhub/jupyter-server-proxy/blame/main/tests/resources/proxyextension.py that for some reason always calls `ProxyHandler.proxy` with `proxied_path = ""` as written in https://github.com/jupyterhub/jupyter-server-proxy/blame/8c5906a68f90252951342ccd626e0332f8531872/tests/resources/proxyextension.py#L36 ...

2xB:
@manics Quick question: I am actively waiting for this since how I implemented another project depends on this at the moment. What is the status here? If there is another way to map paths to websockets, that would be interesting of course as well.

2xB:
@manics If there is anything I can do to support this, please tell.

PR #527, opened 2025-04-22
Add icon type description for the `icon_path` property under ServerProxy.servers

chengs:
in `IconHandler`
https://github.com/jupyterhub/jupyter-server-proxy/blob/8c5906a68f90252951342ccd626e0332f8531872/jupyter_server_proxy/api.py#L71

Reading the icon file is under text (not binary) mode, thus only text-based icon types are supported
by the `icon_path` property. 

The long-term solution would be fixing the `IconHandler` class. 
This PR is a short-team solution, ie, adding description for the `icon_path` mentioning that only text-based 
(well most of cases, SVG) are supported.

### DIFF
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -123,7 +123,7 @@ the following keys:
    explicit entry.
 2. **icon_path**
    Full path to an svg icon that could be used with a launcher. Currently only used by the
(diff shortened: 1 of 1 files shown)


yuvipanda:
Thank you so much for the PR, @chengs! Apologies for the delayed response, we're working on getting better and faster at triage.

PR #532, opened 2025-08-27
Fix openrefine example commands

yuvipanda:
- It's 'refine' not 'openrefine'
- We now need to pass -H * to refine to get it to accept requests from any host, otherwise it will reject based on the Host header

Note that this still prevents it from working with the conda-forge feedstock:
https://github.com/conda-forge/openrefine-feedstock/issues/7


yuvipanda:
Self-merging as a minor doc fix

PR #575, opened 2026-04-17
Support both Unicode and Path for icon_path.

ryanlovett:
Addresses https://github.com/jupyterhub/repo2docker/issues/1520#issuecomment-4270414928

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -16,7 +16,6 @@ else:  # pragma: no cover
 
 from jupyter_server.utils import url_path_join as ujoin
 from traitlets import (
-    Any,
     Bool,
     Callable,
     Dict,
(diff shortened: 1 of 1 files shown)


jupyter_server_proxy/config.py
```diff
@@ -45,14 +46,22 @@ class LauncherEntry(HasTraits):
     """,
     )
 
-    icon_path = Unicode(
-        "",
+    icon_path = Union(
```
manics, line 49: Does `CUnicode` work? It's meant to be a casting variant of `Unicode` so might automatically convert Path to string.
https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.CUnicode

minrk: CUnicode is a little overly permissive because it will accept _anything_ and cast to str. I think we should generally operate as if the `C` casting traitlets don't exist.

PR #577, opened 2026-04-28
fix: add error handling for process instantiation

### LINKED ISSUE #576: Missing error handling for `ServerProcess` construction
When instantiating `ServerProcess` instances after parsing `entrypoints`, there is no error handling to prevent a poorly-behaved plugin from disabling other plugin loading through exception propagation:

https://github.com/jupyterhub/jupyter-server-proxy/blob/2b8a32bbc827be74e7d15499de511c18cea17564/jupyter_server_proxy/config.py#L359

agoose77:
Closes #576 

I did not add any tests, because I created this as a drive-by fix, and didn't have time to figure out the plumbing!

### DIFF
--- a/jupyter_server_proxy/config.py
+++ b/jupyter_server_proxy/config.py
@@ -356,7 +356,13 @@ def get_entrypoint_server_processes():
         except Exception as e:
             warn(f"entry_point {name} was unable to be loaded: {str(e)}")
             continue
-        processes.append(ServerProcess(name=name, **server_process_config))
+        try:
+            process = ServerProcess(name=name, **server_process_config)
+        except Exception as e:
+            warn(f"Process for entry_point {name} could not be instantiated: {str(e)}")
+            continue
+
+        processes.append(process)
     return processes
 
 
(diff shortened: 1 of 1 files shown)


jupyter_server_proxy/config.py
```diff
@@ -356,7 +356,13 @@ def get_entrypoint_server_processes():
         except Exception as e:
             warn(f"entry_point {name} was unable to be loaded: {str(e)}")
             continue
-        processes.append(ServerProcess(name=name, **server_process_config))
+        try:
+            process = ServerProcess(name=name, **server_process_config)
+        except Exception as e:
+            warn(f"Process for entry_point {name} could not be instantiated: {str(e)}")
```
ryanlovett:
```suggestion
            warn(f"Process for entry_point {name} could not be instantiated: {e}")
```
Inside f-string `{}`, `str()` is implicitly called.

ryanlovett (changes requested):
Thanks @agoose77 ! This issue this addresses has actually bugged me for years.

I made one minor suggestions.

jupyter_server_proxy/config.py
agoose77: Ah yes, I've clearly copied the other error block. Let's change this here, and handle the other one down the road.

PR #592, opened 2026-07-16
don't rewrite relative paths in Location

minrk:
only absolute paths need a rewrite (start with /, no scheme)

relative paths should resolve fine, because they are relative and not sensitive to prefixes.

I found this in [jupyter myst build proxy](https://github.com/ryanlovett/jupyter-myst-build-proxy/blob/fa81d1791cd78aa6fde32d1bcdde8f37577df4a8/jupyter_myst_build_proxy/static_server.py#L432) which uses a relative redirect so that it will behave properly under a prefix path, but jupyter server proxy breaks relative redirects, treating them as absolute, turning a redirect from `/prefix/path/to/some/foo` -> `foo/` into `/prefix/foo/` instead of the correct `/prefix/path/to/some/foo/`

### DIFF
--- a/jupyter_server_proxy/handlers.py
+++ b/jupyter_server_proxy/handlers.py
@@ -273,10 +273,14 @@ class ProxyHandler(WebSocketHandlerMixin, JupyterHandler):
         # Parse the location header
         parsed = urlparse(location)
 
-        # Only rewrite if the location is a relative path (no scheme or host)
+        # Only rewrite if the location is an absolute path (no scheme or host, not a relative path)
         if parsed.scheme or parsed.netloc:
             # Absolute URL - leave as is
-            self.log.debug(f"Not rewriting absolute Location header: {location}")
+            self.log.debug(f"Not rewriting full URL Location header: {location}")
+            return location
+        # Relative paths should work file
+        if not location.startswith("/"):
+            self.log.debug(f"Not rewriting relative Location header: {location}")
             return location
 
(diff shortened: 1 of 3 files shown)


ryanlovett:
Thanks for this fix! The `handlers.py` change makes sense.

The change to `redirectserver.py` also affects `test_server_proxy_redirect_location_header_absolute_url` in `test_proxies.py`. That test expects absolute-path Location values but the redirectserver now emits relative paths. Would you be able to fix those tests too?

minrk:
Updated the test so both absolute and relative path redirects get exercised

jupyter_server_proxy/handlers.py
```diff
@@ -273,10 +273,14 @@ def _rewrite_location_header(self, location, host, port, proxied_path):
...
         parsed = urlparse(location)
 
-        # Only rewrite if the location is a relative path (no scheme or host)
+        # Only rewrite if the location is an absolute path (no scheme or host, not a relative path)
         if parsed.scheme or parsed.netloc:
             # Absolute URL - leave as is
-            self.log.debug(f"Not rewriting absolute Location header: {location}")
+            self.log.debug(f"Not rewriting full URL Location header: {location}")
+            return location
+        # Relative paths should work file
```
ryanlovett, line 281: s/file/fine/

ryanlovett (approved):
Looks good, thanks!

PR #595, opened 2026-07-22
Add link to Yuvi's JupyterCon 2025 talk

jasongrout:
### Checklist

- [X] I am the author of this work

### What does this PR do?

Type of change:
- [ ] New feature (non-breaking change which adds functionality)
- [X] Documentation
- [ ] Other

@minrk and I were discussing jupyter-server-proxy the other day, and Min suggested that Yuvi's talk was a really great current reference for jupyter-server-proxy. This PR adds a link on the front page of the documentation.

### Is this PR related to an issue, or is it part of a larger body of work?

No

### Does this PR introduce a breaking change?

No

### How can this PR be tested?

Click the link introduced in the documentation.

### What should a reviewer concentrate their feedback on?

Whether we should link to this talk, and the appropriate place and reference to use.

### DIFF
--- a/docs/source/index.md
+++ b/docs/source/index.md
@@ -21,6 +21,8 @@ The primary use cases are:
    extension](https://github.com/dask/dask-labextension) for
    [dask](https://www.dask.org/).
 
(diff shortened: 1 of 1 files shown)


ryanlovett:
Makes sense, thanks! I checked whether the video could be embedded, as in MyST, but that would need [an extension](https://sphinxcontrib-youtube.readthedocs.io/). I'm not familiar with sphinx, so adding the link seems simplest.

PR #609, opened 2026-09-01
Add exclude_last_activity_patterns to prevent some requests from keeping server alive

minrk:
allows for excluding certain endpoints from activity that might keep the server alive even when it's really idle.

and allow update_last_activity to be a callable

### Checklist

- [x] I am the author of this work

### What does this PR do?

Type of change:
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [ ] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

### How can this PR be tested?

set `exclude_last_activity_patterns` and submit requests. last_activity should not be updated.

### What should a reviewer concentrate their feedback on?

if this is adequate configuration, or if it should be expressed differently

### Other information

cc @shaneknapp who's dealing with lots of idle rstudio sessions not being culled due to constant `POST /rstudio/events/get_events`

I'll do another PR against rsession-proxy to ignore get_events by default

### DIFF
--- a/docs/source/server-process.md
+++ b/docs/source/server-process.md
@@ -161,10 +161,29 @@ One of:
 Whether to report activity from the proxy to Jupyter Server. If _True_, Jupyter Server
 will be notified of new activity. This is primarily used by JupyterHub for idle detection and culling.
 
-Useful if you want to have a seperate way of determining activity through a
+Useful if you want to have a separate way of determining activity through a
 proxied application.
 
 Defaults to _True_.
+
+```{versionchanged} 4.6.0
+`update_last_activity` may be a callable, given the current RequestHandler,
+which should return True or False to track the current handler.
+```
+
+### `exclude_last_activity_patterns`
+
+```{versionadded} 4.6.0
+
+```
+
+Specify a list of regular expressions (patterns as strings, or compiled regular expression objects)
+to _not_ be registered as activity.
+Patterns are tested on the path of the request (e.g. `[".*/proxy/123/some-endpoint$"]` would match `/user/name/proxy/123/some-endpoint`).
+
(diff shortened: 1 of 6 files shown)


shaneknapp:
this is great @minrk!  thanks so much!

ryanlovett (commented):
This will be a useful feature, thanks! A couple comments:

1. In order for downstream proxies to be able to set `exclude_last_activity_patterns` in their configs (like in the draft PR to jupyter-rsession-proxy), it needs to be a ServerProcess traitlet. I think it should be added to `config.py` as a traitlet and then passed through in `get_proxy_attributes`.
2. Should it be documented that when one sets a callable for `update_last_activity`, it will cause `exclude_last_activity_patterns` to be ignored?
3. Maybe tests can be added in a later PR.

minrk:
Added tests and fixed propagation from config, including docstrings explaining that callable `update_last_activity` is the only thing considered, if defined.

ryanlovett:
Thanks, looks good. The `re.compile` in `handlers.py` can probably be removed too, although I think its a no-op if they're already compiled?

minrk:
yeah, should be redundant if we can assume the list always comes via validated config. It is definitely a no-op, `re.compile(re.compile(re.compile('pattern')))` is fine.

minrk:
tests seem to have become flaky, some due to waiting for proxy processes to start on Windows, and some of the unix socket tests occasionally failing with ECONNREFUSED. Not sure what's going on there

minrk:
ok, tests passing after a re-run

ryanlovett:
@minrk I think this okay to merge?

minrk:
Yes, I think so

PR #614, opened 2026-09-02
labextension: update yarn.lock

minrk:
just run jlpm

### Checklist

- [x] I am the author of this work

### What does this PR do?

Type of change:
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [x] Maintenance (testing, packaging, infrastructure, metadata)
- [ ] Other

### DIFF
--- a/labextension/yarn.lock
+++ b/labextension/yarn.lock
@@ -14,30 +14,12 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


minrk:
classic has started failing CI after merging #571 which makes no sense, because that doesn't change any files, and previous success was only seconds earlier. So it seems to be an _extremely_ coincidental time-based change somewhere outside our control.

